PR #30, opened 2016-04-08
RE2 now requires C++11

stefanor:
Since google/re2@cd505f4597d4022902b25bd036de29478e22d481

### DIFF
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -8,7 +8,7 @@ require 'mkmf'
 
 incl, lib = dir_config("re2", "/usr/local/include", "/usr/local/lib")
 
-$CFLAGS << " -Wall -Wextra -funroll-loops"
+$CXXFLAGS << " -Wall -Wextra -funroll-loops"
 
 have_library("stdc++")
 have_header("stdint.h")
@@ -37,7 +37,7 @@ SRC
 
     # Pass -x c++ to force gcc to compile the test program
     # as C++ (as it will end in .c by default).
-    if try_compile(test_re2_match_signature, "-x c++")
+    if try_compile(test_re2_match_signature, "-x c++ #{$CXXFLAGS}")
       $defs.push("-DHAVE_ENDPOS_ARGUMENT")
     end
   end
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -8,7 +8,16 @@ require 'mkmf'
 
 incl, lib = dir_config("re2", "/usr/local/include", "/usr/local/lib")
 
-$CXXFLAGS << " -Wall -Wextra -std=c++11"
+# for try_compile: Pass -x c++ to force gcc to compile the test program as C++
+# (as it will end in .c by default).
+cxx_source = '-x c++'
+
+if defined? $CXXFLAGS
+  $CXXFLAGS << " -Wall -Wextra -std=c++11"
+  cxx_source << " #{$CXXFLAGS}"
+else
+  $CFLAGS << " -Wall -Wextra -std=c++11"
+end
 
 have_library("stdc++")
 have_header("stdint.h")
@@ -35,9 +44,7 @@ int main() {
 }
 SRC
 
-    # Pass -x c++ to force gcc to compile the test program
-    # as C++ (as it will end in .c by default).
-    if try_compile(test_re2_match_signature, "-x c++ #{$CXXFLAGS}")
+    if try_compile(test_re2_match_signature, cxx_source)
       $defs.push("-DHAVE_ENDPOS_ARGUMENT")
     end
   end
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -8,7 +8,7 @@ require 'mkmf'
 
 incl, lib = dir_config("re2", "/usr/local/include", "/usr/local/lib")
 
-$CXXFLAGS << " -Wall -Wextra -funroll-loops -std=c++11"
+$CXXFLAGS << " -Wall -Wextra -std=c++11"
 
 have_library("stdc++")
 have_header("stdint.h")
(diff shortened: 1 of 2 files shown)


mudge:
Thanks for this (and apologies for Hound's noisy auto response, looks like its configuration might need tweaking).

Annoyingly, I can't test this on my ageing computer (Mac OS X 10.7) as it lacks C++11 support. I can test this out elsewhere in a few days though.

Looking at the [official installation instructions](https://github.com/google/re2/wiki/Install), could we simplify the `extconf.rb` by specifically checking for `atomic` (perhaps with `have_library`)?

Also, are the Travis CI configuration changes still required (particularly dropping 1.8.7 support)?

stefanor:
Hound was also right about a few things. I'm not much of a ruby-ist, so I'm not in touch with preferred style :)

> as it lacks C++11 support

Yeah, that's a bit of a downside. There are a lot of systems out there without C++11 support, and requiring it is painful. So, figuring out a way to support both older and newer re2 may be sensible?

> could we simplify the extconf.rb by specifically checking for atomic (perhaps with have_library)?

So, `atomic` is a header in the C++ stdlib. I don't think `have_library` will be that helpful.

> Also, are the Travis CI configuration changes still required

So, look at the first Travis failure. Everything blew up because of lack of C++11 support in Ubuntu 12.04 https://travis-ci.org/mudge/re2/builds/121797286

> (particularly dropping 1.8.7 support)?

I simply couldn't get it to work. https://travis-ci.org/stefanor/ruby-re2/builds/121813125

---

This has all got me thinking that we can probably detect whether the RE2 library requires C++11 yet, or not. And compile, appropriately.

Then you'll be able to revert the travis changes (you're testing against an ancient RE2 library, that doesn't require C++11, anyway). Or keep them, and maybe add a new dimension with a modern RE2.

I'll hack a bit.

stefanor:
There we go. I think that's a lot simpler.

Works with the old Travis config too: https://travis-ci.org/stefanor/ruby-re2/builds/123618713

mudge:
Thanks for that, Stefano.

I'm just trying to test this out locally but the latest stable version of re2 ([2016-04-01](https://github.com/google/re2/releases/tag/2016-04-01)), seems not to require C++11 when linking with this branch:

```
> rake
mkdir -p tmp/x86_64-darwin14/re2/2.3.0
cd tmp/x86_64-darwin14/re2/2.3.0
/Users/mudge/.rubies/ruby-2.3.0/bin/ruby -I. ../../../../ext/re2/extconf.rb
checking for main() in -lstdc++... yes
checking for stdint.h... yes
checking for rb_str_sublen()... yes
checking for RE2... yes
checking for RE2 requires C++11... no
checking for RE2::Match() with endpos argument... yes
creating Makefile
```

Should I be trying to link to a more recent, unreleased version?

ext/re2/extconf.rb
```diff
@@ -8,41 +8,79 @@
 
 incl, lib = dir_config("re2", "/usr/local/include", "/usr/local/lib")
 
-$CFLAGS << " -Wall -Wextra -funroll-loops"
+class CXXFlags
+  def <<(extra_flags)
+    # Newer mkmf
+    if defined? $CXXFLAGS
```
mudge: For posterity: looks like`$CXXFLAGS` was introduced in Ruby 2.2 in https://github.com/ruby/ruby/commit/e8792774dff09386a930f4a93fa6fdc864b2e720

ext/re2/extconf.rb
```diff
@@ -8,41 +8,79 @@
...
+flags << "-Wall -Wextra"
 
 have_library("stdc++")
 have_header("stdint.h")
 have_func("rb_str_sublen")
 
-if have_library("re2")
+checking_for("RE2") do
+  if have_library("re2")
+    true
```
mudge: Could we simplify this by just removing this branch:

```ruby
checking_for("RE2") do
  unless have_library("re2")
    abort "You must have re2 installed and specified with --with-re2-dir, "\
          "please see https://github.com/google/re2/wiki/Install"
  end
end
```

mudge:
Is the new version of re2 exposing any C++11 types or can we get away with linking to it without requiring C++11 on our side (c.f. "[Can a compiled C++11 library (lib,dll,etc.) be linked in older C++ compilers?](http://programmers.stackexchange.com/questions/162408/can-a-compiled-c11-library-lib-dll-etc-be-linked-in-older-c-compilers)")?

ext/re2/extconf.rb
stefanor: Then it reports failure in the output

stefanor:
> Is the new version of re2 exposing any C++11 types

It is using `std::atomic` in `re2.h`, so to consume that header, you need to be using C++11.

stefanor:
> the latest stable version of re2 (2016-04-01), seems not to require C++11

That version should require it. It does for me:

```
mkdir -p tmp/x86_64-linux-gnu/re2/2.3.0
cd tmp/x86_64-linux-gnu/re2/2.3.0
/usr/bin/ruby2.3 -I. ../../../../ext/re2/extconf.rb
checking for main() in -lstdc++... yes
checking for stdint.h... yes
checking for rb_str_sublen()... yes
checking for RE2... yes
checking for RE2 requires C++11... yes
checking for RE2::Match() with endpos argument... yes
creating Makefile
```

mudge:
Just to revive this, I can finally reproduce the build failures and have the library compiling cleanly with Ruby 2.3.1.

To complicate matters, Ruby 2.3.0 had a bug with a use of `memset_s` in `include/ruby/missing.h` which breaks compilation so I'll see if there is some workaround for that version.

mudge:
Thanks so much for this contribution and apologies it took so long to get released but this is now fixed in [version 1.0.0](https://github.com/mudge/re2/releases/tag/v1.0.0).

PR #45, opened 2021-02-06
Update tests.yml

gogainda:


gogainda:
wanted to update bundler version ,but looks like there is more work to do

mudge:
Hi @gogainda,

I've now updated the CI configuration to use Bundler 2 on Ruby 2.1 and newer.

PR #46, opened 2021-02-18
Test TruffleRuby and Ruby 2.7 support

mudge:
Bring the custom re2 test GitHub action into the repository by following https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-action-in-the-same-repository-as-the-workflow, making it easier to manage changes to CI alongside code changes.

By doing so, extend the test action to support versions of TruffleRuby by using the standalone distributions (https://github.com/oracle/truffleruby/blob/master/doc/user/standalone-distribution.md#latest-release).

### DIFF
--- a/.github/actions/test/Dockerfile
+++ b/.github/actions/test/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && \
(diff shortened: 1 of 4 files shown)


mudge:
Going to try a different approach...

PR #56, opened 2022-06-03
Add bindings for RE2::Set

pritambaral:
`RE2::Set` allows for matching multiple patterns against a string. It scans the string only once and simultaneously checks for regexp matches against all patterns added to it. It is thus significantly faster than matching each pattern individually, which would require scanning the string once for every pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,22 @@ enum.next #=> ["It"]
 enum.next #=> ["is"]
 ```
 
+On versions later than 1.4.0, you can use `RE2::Set` to match multiple patterns
+against a string. Calling `RE2::Set#add` with a pattern will return an integer
+index of the pattern. After all patterns have been added, the set can be
+compiled using `RE2::Set#compile`, and then `RE2::Set#match` will return an
+`Array<Integer>` containing the indexes of all the patterns that matched.
+
+``` ruby
+set = RE2::Set.new
+set.add("abc") #=> 0
+set.add("def") #=> 1
+set.add("ghi") #=> 2
+set.compile #=> true
+set.match("abcdefghi") #=> [0, 1, 2]
+set.match("ghidefabc") #=> [2, 1, 0]
+```
+
 Features
 --------
 
--- a/ext/re2/re2.cc
+++ b/ext/re2/re2.cc
@@ -1505,12 +1505,18 @@ static VALUE re2_set_match(VALUE self, VALUE str) {
   Check_Type(str, T_STRING);
   re2::StringPiece data(RSTRING_PTR(str), RSTRING_LEN(str));
   std::vector<int> v;
-  RE2::Set::ErrorInfo e;
   re2_set *s;
   Data_Get_Struct(self, re2_set, s);
+#ifdef HAVE_ERROR_INFO_ARGUMENT
+  RE2::Set::ErrorInfo e;
   bool match_failed = !s->set->Match(data, &v, &e);
+#else
+  bool match_failed = !s->set->Match(data, &v);
+#endif
   VALUE result = rb_ary_new2(v.size());
+
   if (match_failed) {
+#ifdef HAVE_ERROR_INFO_ARGUMENT
     switch (e.kind) {
       case RE2::Set::kNoError:
         break;
@@ -1526,6 +1532,7 @@ static VALUE re2_set_match(VALUE self, VALUE str) {
       default:  // Just in case a future version of libre2 adds new ErrorKinds
         rb_raise(re2_eSetMatchError, "Unknown RE2::Set::ErrorKind: %d", e.kind);
     }
+#endif
   } else {
     for (size_t i = 0; i < v.size(); i++) {
       rb_ary_push(result, INT2FIX(v[i]));
--- a/ext/re2/re2.cc
+++ b/ext/re2/re2.cc
@@ -1443,16 +1443,22 @@ static VALUE re2_set_initialize(int argc, VALUE *argv, VALUE self) {
   } else {
     Check_Type(anchor, T_SYMBOL);
     ID id_anchor = SYM2ID(anchor);
-    if (id_anchor == id_unanchored) { re2_anchor = RE2::UNANCHORED; }
-    else if (id_anchor == id_anchor_start) { re2_anchor = RE2::ANCHOR_START; }
-    else if (id_anchor == id_anchor_both) { re2_anchor = RE2::ANCHOR_BOTH; }
-    else {
+    if (id_anchor == id_unanchored) {
+      re2_anchor = RE2::UNANCHORED;
+    } else if (id_anchor == id_anchor_start) {
+      re2_anchor = RE2::ANCHOR_START;
+    } else if (id_anchor == id_anchor_both) {
+      re2_anchor = RE2::ANCHOR_BOTH;
+    } else {
       rb_raise(rb_eArgError, "anchor should be one of: :unanchored, :anchor_start, :anchor_both");
     }
   }
 
   s->set = new(nothrow) RE2::Set(re2_options, re2_anchor);
-  if (s->set == 0) rb_raise(rb_eNoMemError, "not enough memory to allocate RE2::Set object");
+  if (s->set == 0) {
+    rb_raise(rb_eNoMemError, "not enough memory to allocate RE2::Set object");
+  }
+
   return self;
 }
 
@@ -1472,7 +1478,10 @@ static VALUE re2_set_add(VALUE self, VALUE str) {
   re2_set *s;
   Data_Get_Struct(self, re2_set, s);
   int index = s->set->Add(regex, &err);
-  if (index < 0) rb_raise(rb_eArgError, "str rejected by RE2::Set->Add(): %s", err.c_str());
+  if (index < 0) {
+    rb_raise(rb_eArgError, "str rejected by RE2::Set->Add(): %s", err.c_str());
+  }
+
   return INT2FIX(index);
 }
 
@@ -1503,7 +1512,8 @@ static VALUE re2_set_match(VALUE self, VALUE str) {
   VALUE result = rb_ary_new2(v.size());
   if (match_failed) {
     switch (e.kind) {
-      case RE2::Set::kNoError: break;
+      case RE2::Set::kNoError:
+        break;
       case RE2::Set::kNotCompiled:
         rb_raise(re2_eSetMatchError, "#match must not be called before #compile");
         break;
@@ -1516,10 +1526,10 @@ static VALUE re2_set_match(VALUE self, VALUE str) {
       default:  // Just in case a future version of libre2 adds new ErrorKinds
         rb_raise(re2_eSetMatchError, "Unknown RE2::Set::ErrorKind: %d", e.kind);
     }
-  }
-  else {
-    for(size_t i = 0; i < v.size(); i++)
+  } else {
+    for (size_t i = 0; i < v.size(); i++) {
       rb_ary_push(result, INT2FIX(v[i]));
+    }
   }
   return result;
 }
--- a/ext/re2/re2.cc
+++ b/ext/re2/re2.cc
@@ -8,6 +8,7 @@
 
 #include <ruby.h>
 #include <re2/re2.h>
+#include <re2/set.h>
 #include <stdint.h>
 #include <string>
 #include <sstream>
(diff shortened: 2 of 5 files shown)


mudge:
Thanks for this, @pritambaral. I’m currently on parental leave so can’t promise to thoroughly review this just yet but it sounds like a useful addition.

Please feel free to ignore some of Hound CI’s automatic comments as they are not correct (e.g. the Hash syntax changes it proposes would break compatibility with older Ruby versions).

Has `re2/set` been available since the earliest versions or do we need to detect support at compile time?

pritambaral:
> Has `re2/set` been available since the earliest versions or do we need to detect support at compile time?

Yes, it has been available since the earliest tag I could find on GitHub ([2015-05-21](https://github.com/google/re2/blob/2015-05-01/re2/set.h)], although that version is not API compatible with this PR. The last [API breaking change happened in 2017](https://github.com/google/re2/commit/ee52f030e7fed3fb0cfc5f41c367a898194cf776), so I'm confident that the versions currently supported should work.

spec/re2/set_spec.rb
```diff
@@ -0,0 +1,96 @@
+RSpec.describe RE2::Regexp do
```
mudge, line 1: Should this be `RE2::Set` here?

README.md
```diff
@@ -131,6 +131,22 @@ enum.next #=> ["It"]
...
+compiled using `RE2::Set#compile`, and then `RE2::Set#match` will return an
+`Array<Integer>` containing the indexes of all the patterns that matched.
+
+``` ruby
+set = RE2::Set
+set.add("abc") #=> 0
+set.add("def") #=> 1
+set.add("ghi") #=> 2
+set.compile #=> true
+set.match("abcdefghi") #=> [0, 1, 2]
```
mudge, line 146: This closely matches RE2's own API (`Add`, `Compile` and `Match`) but I wonder if the fact a user has to call `compile` before calling `match` (and that calling `add` after `compile` will raise an error) is surprising in Ruby.

I suppose it's similar to freezing strings (which raises an error if you attempt mutate a string after calling `freeze`) but would it be worth having separate classes for uncompiled Sets and compiled Sets, e.g.

```ruby
set = RE2::Set.new
#=> #<RE2::Set ...>
set.add("foo")
set.match("foooo")
#=> NoMethodError: undefined method `match' for #<RE2::Set ...>
set.compile
#=> #<RE2::CompiledSet ...>
set.add("foo")
#=> NoMethodError: undefined method `add' for #<RE2::CompiledSet ...>
set.match("fooooo")
#=> [0, 1, 2]
```

This way we're enforcing the correct usage at the API level rather than with runtime errors. That said, perhaps the runtime errors are more informative?

pritambaral: I don't think closely matching RE2's own API is too surprising in Ruby, but then again I don't have a lot of experience writing Ruby code, only reading.

However, having separate classes like you suggest seems, to me, more surprising. I'm assuming you meant that `Set#compile` returns a new `CompiledSet` but does not change `self` from an object of `Set` into an object of `CompiledSet`, since Ruby provides no way (that I could find) to change the class of an object. This raises two issues:

1. `RE2::Set` is now no longer an `RE2::Set`, since you can't actually use it to match regexps. At best, it is sth more like an `RE2::SetBuilder`.
2. Even if we rename `RE2::Set` to `RE2::SetBuilder` (or sth like that), its utility would be immediately lost when `#compile` is called on it. Or, we make it capable of adding new patterns after `#compile`, so a new `CompiledSet` may be returned from a subsequent call to `#compile` that includes both the initial and subsequent patterns.
3. If we were to add the ability to generate multiple `CompiledSet`s from a single `SetBuilder`, we'd need to save every pattern that's been added to a `SetBuilder` instance, and within every call to `#compile`, create a new C++ `RE2::Set` object internally and re-add all previous patterns to it. Any call to `#compile` at any time could then compile this internal C++ `RE2::Set` and return it.

But, that doesn't seem very idiomatic Ruby anyway, and there exists a better option:

```ruby
module RE2
  # @example
  #    set = RE2::Set(["abc", "def", "ghi"])
  #    set.match("abcdefghi") => [0, 1, 2]
  def Set(patterns, anchor = :unanchored, options = {})
    set = RE2::Set.new(anchor, options)
    patterns.each { |p| set.add(p) }
    set.compile
    set
  end
end
```

----

My best reason to stick to C++ RE2's own API is that it results in a clean mapping for users coming from C++ (like me). My second best reason is that I already have code written in this style. Of course, I have no issues against making a top-level function that takes a list of patterns to add and compile in a single-shot. But I'd prefer to not hide the internals for anyone else like me to also be able to use it, if they so please.

README.md
```diff
@@ -131,6 +131,22 @@ enum.next #=> ["It"]
...
 ```
 
+On versions later than 1.4.0, you can use `RE2::Set` to match multiple patterns
+against a string. Calling `RE2::Set#add` with a pattern will return an integer
+index of the pattern. After all patterns have been added, the set can be
+compiled using `RE2::Set#compile`, and then `RE2::Set#match` will return an
+`Array<Integer>` containing the indexes of all the patterns that matched.
+
+``` ruby
+set = RE2::Set
```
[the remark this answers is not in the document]
pritambaral: Of course.

spec/re2/set_spec.rb
pritambaral: Of course. Thanks.

spec/re2/set_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  # :log_errors is true. So we need to temporarily redirect stderr while running
+  # tests that might print to it. The method used here is taken from the source
+  # code of MRI and seems to do the job, but it isn't guaranteed to cover all
+  # cases.
+
+  NullDevice = defined?(IO::NULL) ? IO::NULL :
+    %w[/dev/null NUL NIL: NL:].find {|dev| File.exist?(dev)}
+
+  def silence_streams(*streams)
+    on_hold = streams.collect { |stream| stream.dup }
```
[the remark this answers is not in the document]
pritambaral: I'm ignoring this because I don't know if the suggestions would break on older versions of Ruby, and I don't really want to test it.

README.md
mudge: That makes sense and your point about a `RE2::SetBuilder` becoming useless after calling `compile` (without reworking its internals to manage multiple underlying `RE2::Set`s) is a good one. I was hoping to avoid the mutable state of a single instance by having a more functional interface but this shows we can't really escape the need for one object to become "compiled" and therefore unusable.

Keeping things similar to the underlying C++ API seems like the best course of action.

spec/re2/set_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    streams.each do |stream|
+      stream.reopen(NullDevice)
+      stream.sync = true
+    end
+    yield
+  ensure
+    streams.each_with_index do |stream, i|
+      stream.reopen(on_hold[i])
+    end
+  end
```
mudge, line 22: Rather than reopening `STDERR` to a platform-specific null device, could we take [a leaf from RSpec](https://github.com/rspec/rspec-expectations/blob/43bf64b01f8356979ffbc373b2e81d2ab1389b29/lib/rspec/matchers/built_in/output.rb#L159-L176) and re-assign `$stderr` to a disposable `StringIO` we let Ruby garbage collect without ever reading? e.g.

```ruby
def silence_stderr(&blk)
  null_stream = StringIO.new
  original_stream = $stderr
  $stderr = null_stream

  blk.call
ensure
  $stderr = original_stream
end

# ...

silence_stderr do
  expect { set.add("def") }.to raise_error(ArgumentError)
end
```

ext/re2/re2.cc
```diff
@@ -1362,6 +1375,152 @@ static VALUE re2_QuoteMeta(VALUE self, VALUE unquoted) {
...
+ * called after #compile has been called.
+ *
+ * @param [String] str the regex pattern
+ * @return [Integer] the index of the pattern in the set
+ * @example
+ *   RE2::Set.new.add("abc")    #=> 0
+ */
+static VALUE re2_set_add(VALUE self, VALUE str) {
+  if(TYPE(str) != T_STRING) rb_raise(rb_eArgError, "str must be a string");
+  re2::StringPiece regex(RSTRING_PTR(str), RSTRING_LENINT(str));
```
mudge, line 1470: Is `RSTRING_LENINT` available in older versions of Ruby? If not, we currently use `static_cast<int>(RSTRING_LEN(str))` elsewhere.

spec/re2/set_spec.rb
```diff
@@ -0,0 +1,96 @@
+RSpec.describe RE2::Set do
+  # libre2, as of 2022-06-01, in set.cc, does print to stderr (fd 2) even when
+  # :log_errors is true. So we need to temporarily redirect stderr while running
```
mudge, line 3: Should this be `:log_errors is false`?

spec/re2/set_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+  end
+
+  describe "#add" do
+    it "allows multiple patterns to be added" do
+      set = RE2::Set.new
+      expect(set.add("abc")).to eq(0)
+      expect(set.add("def")).to eq(1)
+      expect(set.add("ghi")).to eq(2)
+    end
```
mudge, line 51: If we have multiple expectations in a single example, it might be worth using RSpec's aggregate failures to combine them into one, e.g.

```suggestion
    it "allows multiple patterns to be added", :aggregate_failures do
      set = RE2::Set.new
      expect(set.add("abc")).to eq(0)
      expect(set.add("def")).to eq(1)
      expect(set.add("ghi")).to eq(2)
    end
```

spec/re2/set_spec.rb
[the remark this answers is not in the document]
mudge: I've approved the CI workflow so any changes to this PR should now run the full test suite.

mudge:
I've now removed the Hound webhook so it should no longer add noise to the review.

ext/re2/re2.cc
```diff
@@ -1362,6 +1375,152 @@ static VALUE re2_QuoteMeta(VALUE self, VALUE unquoted) {
...
+void re2_set_free(re2_set *self) {
+  if (self->set) {
+    delete self->set;
+  }
+  free(self);
+}
+
+static VALUE re2_set_allocate(VALUE klass) {
+  re2_set *s;
+  VALUE result = Data_Make_Struct(klass, re2_set, 0, re2_set_free, s);
```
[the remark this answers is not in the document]
pritambaral: IIUC, `rb_gc_mark` is to be used only on Ruby data internal to `Data_Struct` classes. `self->set` is pure C++ data type. The `re2_set` struct has no `VALUE`-typed members, so we don't need to mark anything internally.

ext/re2/re2.cc
```diff
@@ -1362,6 +1375,152 @@ static VALUE re2_QuoteMeta(VALUE self, VALUE unquoted) {
...
+    re2_anchor = RE2::UNANCHORED;
+  } else {
+    Check_Type(anchor, T_SYMBOL);
+    ID id_anchor = SYM2ID(anchor);
+    if (id_anchor == id_unanchored) re2_anchor = RE2::UNANCHORED;
+    else if (id_anchor == id_anchor_start) re2_anchor = RE2::ANCHOR_START;
+    else if (id_anchor == id_anchor_both) re2_anchor = RE2::ANCHOR_BOTH;
+    else {
+      rb_raise(rb_eArgError, "anchor should be one of: :unanchored, :anchor_start, :anchor_both");
+    }
```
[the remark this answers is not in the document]
pritambaral: > use a `switch`?

`switch` requires `case` predicates to be compile-time constants. I tried but found no way to make the ids `const`, since `rb_intern()` cannot be a `constexpr`-fn.

> could we prefer conditionals with braces

This shall be done.

ext/re2/re2.cc
```diff
@@ -1362,6 +1375,152 @@ static VALUE re2_QuoteMeta(VALUE self, VALUE unquoted) {
...
+ * Adds a regex pattern to the set. Returns index of added pattern. Cannot be
+ * called after #compile has been called.
+ *
+ * @param [String] str the regex pattern
+ * @return [Integer] the index of the pattern in the set
+ * @example
+ *   RE2::Set.new.add("abc")    #=> 0
+ */
+static VALUE re2_set_add(VALUE self, VALUE str) {
+  if(TYPE(str) != T_STRING) rb_raise(rb_eArgError, "str must be a string");
```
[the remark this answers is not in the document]
pritambaral: I didn't know if `Check_Type` is available in older versions of Ruby, so I checked for existing type checks in this file and found all of them to be manual checks like this.

It appears I missed one instance (in `re2_set_initialize`) and accidentally slipped in a `Check_Type`. On further spelunking into Ruby's source repo, it appears okay to use `Check_Type` on older versions of Ruby. I shall change this (and below) accordingly.

ext/re2/re2.cc
pritambaral: It appears to have been introduced in 1.9. Thanks for this catch; will change accordingly.

spec/re2/set_spec.rb
pritambaral: I wasn't sure if simply reassigning a ruby variable would also change any libc-level `stderr` global variable, so I went looking at how MRI itself does this. I've now tested this snippet from RSpec and verified it works, so will change it.

ext/re2/re2.cc
```diff
@@ -1362,6 +1375,152 @@ static VALUE re2_QuoteMeta(VALUE self, VALUE unquoted) {
...
+  return BOOL2RUBY(s->set->Compile());
+}
+
+/*
+ * @param [String] str the text to match against
+ * @return [Array<Integer>] the indices of matching regexps
+ */
+static VALUE re2_set_match(VALUE self, VALUE str) {
+  if(TYPE(str) != T_STRING) rb_raise(rb_eArgError, "str must be a string");
+  VALUE result = rb_ary_new();
```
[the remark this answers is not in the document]
pritambaral: Could, will.

ext/re2/re2.cc
```diff
@@ -93,12 +94,85 @@ typedef struct {
...
+  if (!NIL_P(word_boundary)) {
+    re2_options.set_word_boundary(RTEST(word_boundary));
+  }
+
+  one_line = rb_hash_aref(options, ID2SYM(id_one_line));
+  if (!NIL_P(one_line)) {
+    re2_options.set_one_line(RTEST(one_line));
+  }
+
+  return re2_options;
```
mudge, line 174: The extraction of this function seems to be causing an issue with versions of re2 with an ABI version of 0, see https://github.com/mudge/re2/runs/7521387298?check_suite_focus=true

```
/usr/include/re2/re2.h: In function ‘re2::RE2::Options parse_re2_options(VALUE)’:
/usr/include/re2/re2.h:675:5: error: ‘re2::RE2::Options::Options(const re2::RE2::Options&)’ is private
     Options(const Options&);
     ^
../../../../ext/re2/re2.cc:174:10: error: within this context
   return re2_options;
          ^
In file included from ../../../../ext/re2/re2.cc:10:0:
/usr/include/re2/re2.h: In function ‘VALUE re2_regexp_initialize(int, VALUE*, VALUE)’:
/usr/include/re2/re2.h:675:5: error: ‘re2::RE2::Options::Options(const re2::RE2::Options&)’ is private
     Options(const Options&);
     ^
../../../../ext/re2/re2.cc:751:57: error: within this context
     RE2::Options re2_options = parse_re2_options(options);
                                                         ^
In file included from ../../../../ext/re2/re2.cc:10:0:
/usr/include/re2/re2.h: In function ‘VALUE re2_set_initialize(int, VALUE*, VALUE)’:
/usr/include/re2/re2.h:676:10: error: ‘void re2::RE2::Options::operator=(const re2::RE2::Options&)’ is private
     void operator=(const Options&);
          ^
../../../../ext/re2/re2.cc:1437:17: error: within this context
     re2_options = parse_re2_options(options);
                 ^
In file included from ../../../../ext/re2/re2.cc:10:0:
/usr/include/re2/re2.h:676:10: error: ‘void re2::RE2::Options::operator=(const re2::RE2::Options&)’ is private
     void operator=(const Options&);
          ^
../../../../ext/re2/re2.cc:1439:17: error: within this context
     re2_options = RE2::DefaultOptions;
```

spec/re2/set_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      set = RE2::Set.new
+      set.add("abc")
+      set.add("def")
+      set.add("ghi")
+      set.compile
+      expect(set.match("abcdefghi")).to eq([0, 1, 2])
+    end
+
+    it "raises an error if called before #compile" do
+      pending "Versions of re2 with ABI 0 will only return false here"
```
mudge, line 113: This behaviour discrepancy between underlying versions of re2 makes me uneasy: if you have a modern version of the library, `#match` will raise an exception but if you have an old version, you'll get a `false` return value and no error.

My concern here is that a change in the underlying re2 library will cause significant change in any client programs, especially as a raised exception will alter the control flow.

However, it doesn't seem trivial to backfill this behaviour to re2 with an ABI version of 0 as not enough information is exposed by `RE2::Set::Match()` there.

mudge (changes requested):
Apologies for the delay, @pritambaral (I am still on leave).

I've managed to get the CI build up and running again by updating to a supported runner and recompiling libre2-dev for Ubuntu 20.04. This has exposed a few issues with older versions of the underlying re2 library that it'd be good to get your opinion on.

I've made a start at compatibility for ABI version 0 but there is a tricky API issue re the lack of meaningful error information in `RE2::Set::Match()`.

spec/re2/set_spec.rb
mudge: You were right about this: I switched back to reopening `STDERR` with a platform-specific null device in ecbeae6c00fdf912fe993924253c981496fd3d2e.

mudge: You were right about this: I've reverted back to reopening `STDERR` with the platform-specific null device in ecbeae6c00fdf912fe993924253c981496fd3d2e.

ext/re2/re2.cc
mudge: Fixed in d6252548ab6ff39df35e56042df9d41ef99a44ac by passing `RE2::Options` by reference and mutating them in place rather than attempting to return a new instance.

spec/re2/set_spec.rb
mudge: I've updated this in c00eda1385c722b1063a7d33d26cb6e38b06f4b8  so there is now a `RE2::Set.match_raises_errors?` predicate that returns `true` or `false` based on the underlying re2 library's support. We can then distinguish the two behaviours: raises errors or simply returning an empty array of matches.

I'm still not decided on the final API here though. Should calling `RE2::Set#match` raise or not raise errors depending on the value of `RE2::Set.match_raises_errors?` or should we take a leaf from [`Integer`](https://ruby-doc.org/core-3.1.2/Kernel.html#Integer-method) and have an `exception` argument that defaults to `true` and raises a [`NotImplementedError`](https://ruby-doc.org/core-3.1.2/NotImplementedError.html) if the user tries to use that against a version of re2 that doesn't output error information?

```ruby
set = RE2::Set.new
set.add("foo")
set.add("bar")
set.match("foobar")
# NotImplementedError "current version of RE2::Set::Match does not output error information"
set.match("foobar", exception: false)
# => []
```

mudge: I implemented the `:exception` option in ebfe0336dc6fed98496e8141aa992ad0e6a2dbb0.

mudge:
I've cleaned up the history and rolled these changes into [1.5.0](https://github.com/mudge/re2/releases/tag/v1.5.0) which should now be [available from RubyGems](https://rubygems.org/gems/re2/versions/1.5.0).

Thank you for your contribution and extraordinary patience.

Issue #61, opened 2022-12-02
Vendor re2


mudge:
As suggested by @tomstuart: so that users don’t need to separately install the underlying re2 library, provide a vendored version that is compiled when the gem is installed.

In order to avoid breaking compatibility for existing users, the system re2 should still be preferred (either automatically found in `/usr/local`, `/opt/homebrew` or `/usr` or specified during install with `—with-re2-dir`) with the vendored version only built and used as a fallback.

## Starting points

* https://github.com/libgit2/rugged/blob/56db1abdc69a29627045d0adb53588ba13d9546e/ext/rugged/extconf.rb
* https://github.com/sparklemotion/nokogiri/blob/6d2c1fa5064c9b599618beb23aef1d2fed172eda/ext/nokogiri/extconf.rb
* https://github.com/Homebrew/homebrew-core/blob/047fb73de4efb8bad1b743d91ffcf2fd382045cc/Formula/re2.rb#L38

## Questions

* How can we reliably compile our own version of re2 that links to the gem successfully?
* How do we support various different platforms (including Windows)?
* What are the licensing implications of vendoring the upstream re2 source code inside the gem?
* Does this put pressure on releasing the gem every time there is an re2 update?

stanhu:
I came across this issue because every time Homebrew updates `re2`, this appears to break the gem and users have to run `gem pristine re2` to get back into a working state. I was looking to make this gem statically link `re2`, but it appears that Homebrew only ships the shared library. https://github.com/Homebrew/homebrew-core/pull/119928 fixes that.

> How can we reliably compile our own version of re2 that links to the gem successfully?

https://github.com/flavorjones/mini_portile, which Nokogiri uses, should help. I think you also have to keep in mind to include the `.tar.gz` with the gem because you'll want to avoid downloading the source from an external site every time the gem installs.

Related wotk with ruby-magic:

1. https://github.com/kwilczynski/ruby-magic/pull/3
1. https://github.com/kwilczynski/ruby-magic/pull/5

>  How do we support various different platforms (including Windows)?

mini_portile should You'll see Nokogiri has some special handling for Windows.

> What are the licensing implications of vendoring the upstream re2 source code inside the gem?

I suspect there's no issue since the re2 is a BSD license, and you're not actually modifying it?

> Does this put pressure on releasing the gem every time there is an re2 update?

Perhaps. It looks like re2 updates every few months.

mudge:
[...]

PR #64, opened 2023-01-06
Support static linking of re2 library

stanhu:
Currently when the re2 system library is updated by Homebrew, the dynamically linked library is removed and moved to a new location. This breaks the re2 gem and requires users to run `gem pristine re2` to get back into a working state.

To avoid this hassle, this commit adds a `--enable-static` flag that will statically link the library in the gem's C extension library.

Homebrew doesn't currently install a static library for re2. This will be fixed via https://github.com/Homebrew/homebrew-core/pull/119928.

NOTE: This doesn't work at the moment with Ubuntu's `libre2-dev` package since a number of objects aren't compiled with `-fPIC`: https://stackoverflow.com/a/4036497.

### DIFF
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -5,6 +5,7 @@
 # Released under the BSD Licence, please see LICENSE.txt
 
 require 'mkmf'
+require 'shellwords'
 
 if ENV["CC"]
   RbConfig::MAKEFILE_CONFIG["CC"] = ENV["CC"]
@@ -22,13 +23,40 @@ header_dirs = [
   "/usr/include"
 ]
 
-lib_dirs = [
+LIB_DIRS = [
   "/usr/local/lib",
   "/opt/homebrew/lib",
-  "/usr/lib"
-]
+  "/usr/lib",
+  "/usr/lib/x86_64-linux-gnu"
+].select { |dir| Dir.exist?(dir) }
 
-dir_config("re2", header_dirs, lib_dirs)
+def libflag_to_filename(ldflag)
+  case ldflag
+  when /\A-l(.+)/
+    "lib#{Regexp.last_match(1)}.#{$LIBEXT}"
+  end
+end
+
+def resolve_static_library(arg)
+  libs_path = pkg_config('re2', 'libs-only-L')
+
+  re2_dirs =
+    if libs_path.nil? || libs_path.empty?
+      LIB_DIRS
+    else
+      # There should only be one path, but assume there can be multiple.
+      libs_path.strip.split(' ').map { |lib| lib.gsub(/^-L/, '') }
+    end
+
+  filename = libflag_to_filename(arg)
+  dir = re2_dirs.find { |path| File.exist?(File.join(path, filename)) }
+
+  raise "Unable to find #{filename} in #{re2_dirs}" unless dir
+
+  File.join(dir, filename)
+end
+
+dir_config("re2", header_dirs, LIB_DIRS)
 
 $CFLAGS << " -Wall -Wextra -funroll-loops"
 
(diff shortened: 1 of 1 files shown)


ext/re2/extconf.rb
```diff
@@ -28,6 +29,28 @@
...
 
+def libflag_to_filename(ldflag)
+  case ldflag
+  when /\A-l(.+)/
+    "lib#{Regexp.last_match(1)}.#{$LIBEXT}"
+  end
+end
+
+def resolve_static_library(arg)
+  libs_path = pkg_config('re2', 'libs-only-L')
```
stanhu, line 40: On Linux, this appears to return an empty list.

stanhu: It looks like the static library isn't shipped in the `libre2-5` package.

stanhu: However, the static library is shipped in `libre2-dev` under `/usr/lib/x86_64-linux-gnu/libre2.a`.

stanhu: Hmm, still getting:

```
g++ -shared -o re2.so re2.o -L. -L/home/stanhu/.rbenv/versions/2.7.5/lib -Wl,-rpath,/home/stanhu/.rbenv/versions/2.7.5/lib -L/usr/local/lib -Wl,-rpath,/usr/local/lib -L/opt/homebrew/lib -Wl,-rpath,/opt/homebrew/lib -L/usr/lib -Wl,-rpath,/usr/lib -L/usr/lib/x86_64-linux-gnu -Wl,-rpath,/usr/lib/x86_64-linux-gnu -L. -L/home/stanhu/.rbenv/versions/2.7.5/lib  -fstack-protector-strong -rdynamic -Wl,-export-dynamic -L/home/stanhu/.rbenv/versions/2.7.5/lib  -Wl,--compress-debug-sections=zlib    -Wl,-rpath,/home/stanhu/.rbenv/versions/2.7.5/lib -L/home/stanhu/.rbenv/versions/2.7.5/lib -lruby /usr/lib/x86_64-linux-gnu/libre2.a -lstdc++ -lm   -lc
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libre2.a(re2.o): relocation R_X86_64_PC32 against symbol `_ZTVSt9basic_iosIcSt11char_traitsIcEE@@GLIBCXX_3.4' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
```

stanhu: Could be related to Ubuntu's compilation: https://github.com/google/re2/issues/178

stanhu: Perhaps we're better off doing https://github.com/mudge/re2/issues/61.

stanhu: Confirmed that the installed Ubuntu shared library appears to be missing `-fPIC` in some objects. Recompiling with `CPPFLAGS=-fPIC make` seems to work.

mudge:
Can you explain at a high-level how you statically link a library in a Ruby C extension? It looks like you’re replacing the `-lre2` flag in `$libs` with the absolute path of a specific re2 file?

stanhu:
> Can you explain at a high-level how you statically link a library in a Ruby C extension? It looks like you’re replacing the -lre2 flag in $libs with the absolute path of a specific re2 file?

Yes. That's what Nokogiri does: https://github.com/sparklemotion/nokogiri/blob/b57c92686fbdb90e3e3947ceff654ae825caa2b9/ext/nokogiri/extconf.rb#L975-L986

mudge:
Would it be possible to add a static install to the CI build matrix? It’d be good to see how it behaves against an re2 install that has a static library as well as one that doesn’t. You can see  https://github.com/mudge/re2-ci/blob/853bf95e8e099a91eb9bf06f2b02b2626129e01b/libre2-dev/16.04/entrypoint.sh#L10 for how the various re2 debs are built for GitHub Actions.

stanhu:
@mudge At the moment, this doesn't work with Debian/Ubuntu because the re2 static library isn't compiled with `-fPIC`.

I think I'll try to implement https://github.com/mudge/re2/issues/61 and then add this option there so that we have more control over how the library is built.

stanhu:
Closing in favor of #67.

PR #65, opened 2023-07-03
Fix compilation with re2 2023-07-01

stanhu:
https://code-review.googlesource.com/c/re2/+/61250 made re2 depend directly on abseil. The latest version of abseil now requires C++14 (https://github.com/abseil/abseil-cpp/issues/1127). However, it seems that supplying `-std=c++14` isn't enough (https://github.com/abseil/abseil-cpp/issues/1431). `-std=c++17` needs to be used at least.

This commit fixes the compilation by trying C++20 and C++17 if the initial compilation fails. This fixes build issues on macOS.

### DIFF
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -49,39 +49,27 @@ minimal_program = <<SRC
 int main() { return 0; }
 SRC
 
-success = false
-
-checking_for("re2 requires explicit C++ version flag") do
-  if try_compile(minimal_program, compile_options)
-    success = true
-    false
-  else
-    true
-  end
+re2_requires_version_flag = checking_for("re2 that requires explicit C++ version flag") do
+  !try_compile(minimal_program, compile_options)
 end
 
-unless success
-  TRY_CXXFLAGS = %w[-std=c++20 -std=c++17 -std=c++11 -std=c++0x].freeze
-
+if re2_requires_version_flag
   # Recent versions of re2 depend directly on abseil, which requires a
   # compiler with C++14 support (see
   # https://github.com/abseil/abseil-cpp/issues/1127 and
   # https://github.com/abseil/abseil-cpp/issues/1431). However, the
   # `std=c++14` flag doesn't appear to suffice; we need at least
   # `std=c++17`.
-  TRY_CXXFLAGS.each do |version_flag|
-    checking_for("re2 compiles with #{version_flag}") do
-      if try_compile(minimal_program, compile_options + " #{version_flag}")
-        compile_options << " #{version_flag}"
-        $CPPFLAGS << " #{version_flag}"
-        success = true
+  abort "Cannot compile re2 with your compiler: recent versions require C++14 support." unless %w[c++20 c++17 c++11 c++0x].any? do |std|
+    checking_for("re2 that compiles with #{std} standard") do
+      if try_compile(minimal_program, compile_options + " -std=#{std}")
+        compile_options << " -std=#{std}"
+        $CPPFLAGS << " -std=#{std}"
+
+        true
       end
     end
-
-    break success if success
   end
-
-  abort "Cannot compile re2 with your compiler: recent versions require C++14 support." unless success
 end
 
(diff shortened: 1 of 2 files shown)


stanhu:
It seems `abseil` is no longer an optional dependency on a separate branch: https://github.com/google/re2/issues/388

https://code-review.googlesource.com/c/re2/+/61270 updated the `README.md` to make this official.

mudge:
Thanks for this, @stanhu. I'll try and get the new 2023-07-01 release (with its added abseil dependency) into the CI matrix (via [re2-ci](https://github.com/mudge/re2-ci/tree/main/libre2-dev)).

ext/re2/extconf.rb
```diff
@@ -44,24 +44,35 @@
...
-checking_for("re2 requires C++11 compiler") do
+TRY_CXXFLAGS = %w[-std=c++20 -std=c++17 -std=c++11 -std=c++0x]
+
+# Recent versions of re2 depend directly on abseil, which requires a
+# compiler with C++14 support (see
+# https://github.com/abseil/abseil-cpp/issues/1127 and
+# https://github.com/abseil/abseil-cpp/issues/1431). However, the
+# `std=c++14` flag doesn't appear to suffice; we need at least
+# `std=c++17`.
+checking_for("re2 requires a C++14 compiler") do
```
mudge, line 55: Is there a way for us to print the relevant message as we check for a C++17 compiler separately from checking for a C++11 compiler (for versions of re2 prior to release 2023-07-01)?

It's also worth noting that the return value of the block is used to determine whether it will print `yes` or `no` when compiling, e.g. when I tried this on my machine with the latest release, I see:

```
checking for re2 requires a C++14 compiler... no
```

stanhu: @mudge Good point, I've improved the messages.

mudge:
Have you had any luck compiling the latest re2 from scratch on Ubuntu 20.04?

stanhu:
> Have you had any luck compiling the latest re2 from scratch on Ubuntu 20.04?

@mudge Yes, but only by [installing a PPA with `libabsl-dev`](https://launchpad.net/~savoury1/+archive/ubuntu/build-tools. It seems that the `libabsl-dev` package is only available with Ubuntu 22.04: https://packages.ubuntu.com/jammy/libabsl-dev. This is what I ran:

```
docker run -it ubuntu:20.04 bash
apt update && apt install -y software-properties-common build-essential pkg-config git
add-apt-repository ppa:savoury1/build-tools
apt install -y libabsl-dev
git clone https://github.com/google/re2.git /tmp/re2
cd /tmp/re2
make
```

mudge:
I hoped I could bump the more recent Ruby builds to Ubuntu 22.04 so we could use `libabsl-dev` (and `libabsl20210324`) but it looks like there are [fatal problems trying to use RubyGems on Ubuntu 22.04](https://github.com/ruby/setup-ruby/issues/496) so I'll need to use the PPA or package up my own `libabsl` and stick with Ubuntu 20.04 for older Rubies.

stanhu:
@mudge Is there a reason these older Ruby versions still need to be supported? Even Ruby 2.7 reached end-of-life 3 months ago.

mudge:
It's not a popular opinion but until it becomes extraordinarily painful (which might not be too far away), I don't want to break compatibility for anyone who can already run the gem.

In this case, I've added the PPA you mentioned to the existing 20.04 build (and I've packaged up a `.deb` of re2 2023-07-01 compiled against the `libabsl-dev` from the PPA) but it looks like the build is failing: https://github.com/mudge/re2/actions/runs/5456057642/jobs/9928355646

mudge:
OK, I've fixed the build (`libabsl2206` isn't enough, we needed the full `libabsl-dev`).

mudge:
Thanks @stanhu, I've now rebased this into https://github.com/mudge/re2/commit/8b62c5c4ac5552d6b3465d0d4a3b6494604bcb1b and it'll go out in version 1.7.0 once the tests pass on `main`.

PR #67, opened 2023-07-11
Vendor re2 and abseil

stanhu:
This pull request:

* Vendors re2 and abseil in the Ruby platform gem.
* Builds precompiled native gems for the following platforms:

  - aarch64-linux
  - arm-linux
  - arm64-darwin
  - x64-mingw-ucrt
  - x86-linux
  -  x86-mingw32
  - x86_64-darwin
  - x86_64-linux

* By default, build and link against the vendored re2 and abseil versions. `cmake` and a C++17 compiler is required for this to work. Users can specify `--enable-system-libraries` or set `RE2_USE_SYSTEM_LIBRARIES`.

This makes it possible to ensure all the required dependencies are contained within the C extension rather than depend on system libraries, which may break the extension when they are updated.

The building of these libraries uses mini_portile2 and techniques borrowed from the nokogiri and ruby-magic gems.

glibc 2.29 is required for the arm64 Linux installs.

Closes https://github.com/mudge/re2/issues/61

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -62,7 +62,7 @@ namespace 'gem' do
       RakeCompilerDock.sh <<~SCRIPT, platform: platform, verbose: true
         gem install bundler --no-document &&
         bundle &&
-        CMAKE=#{cmake} bundle exec rake gem:#{platform}:builder MAKE='nice make -j`nproc`'
+        bundle exec rake gem:#{platform}:builder CMAKE=#{cmake}
       SCRIPT
     end
 
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,17 @@
 require 'rake/extensiontask'
 require 'rspec/core/rake_task'
 
+CLEAN.include FileList['**/*{.o,.so,.dylib,.bundle}'],
+              FileList['**/extconf.h'],
+              FileList['**/Makefile'],
+              FileList['pkg/']
+
+CLOBBER.include FileList['**/tmp'],
+                FileList['**/*.log'],
+                FileList['doc/**'],
+                FileList['tmp/']
+CLOBBER.add("ports/*").exclude(%r{ports/archives$})
+
 Rake::ExtensionTask.new('re2')
 
 RSpec::Core::RakeTask.new(:spec)
--- a/Rakefile
+++ b/Rakefile
@@ -25,7 +25,11 @@ end
 CROSS_RUBY_VERSIONS = %w[3.2.0 3.1.0 3.0.0 2.7.0].join(':')
 CROSS_RUBY_PLATFORMS = %w[
   aarch64-linux
+  arm-linux
   arm64-darwin
+  x64-mingw-ucrt
+  x86-linux
+  x86-mingw32
   x86_64-darwin
   x86_64-linux
 ].freeze
@@ -56,7 +60,14 @@ namespace 'gem' do
     # The Linux x86 image (ghcr.io/rake-compiler/rake-compiler-dock-image:1.3.0-mri-x86_64-linux)
     # is based on CentOS 7 and has two versions of cmake installed:
     # a 2.8 version in /usr/bin and a 3.25 in /usr/local/bin. The latter is needed by abseil.
-    cmake = platform == 'x86_64-linux' ? '/usr/local/bin/cmake' : 'cmake'
+    cmake =
+      case platform
+      when 'x86_64-linux', 'x86-linux'
+        '/usr/local/bin/cmake'
+      else
+        'cmake'
+      end
+
     desc "build native gem for #{platform} platform"
     task platform do
       RakeCompilerDock.sh <<~SCRIPT, platform: platform, verbose: true
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,8 @@
+# frozen_string_literal: true
+
 require 'rake/extensiontask'
 require 'rspec/core/rake_task'
+require 'rake_compiler_dock'
 
 CLEAN.include FileList['**/*{.o,.so,.dylib,.bundle}'],
               FileList['**/extconf.h'],
@@ -14,17 +17,65 @@ CLOBBER.add("ports/*").exclude(%r{ports/archives$})
 
 RE2_GEM_SPEC = Gem::Specification.load('re2.gemspec')
 
-Rake::ExtensionTask.new('re2')
-
 Gem::PackageTask.new(RE2_GEM_SPEC) do |p|
   p.need_zip = false
   p.need_tar = false
 end
 
+CROSS_RUBY_VERSIONS = %w[3.2.0 3.1.0 3.0.0 2.7.0].join(':')
+CROSS_RUBY_PLATFORMS = %w[
+  aarch64-linux
+  arm64-darwin
+  x86_64-darwin
+  x86_64-linux
+].freeze
+
+ENV['RUBY_CC_VERSION'] = CROSS_RUBY_VERSIONS
+
+Rake::ExtensionTask.new('re2', RE2_GEM_SPEC) do |e|
+  e.cross_compile = true
+  e.cross_config_options << '--enable-cross-build'
+  e.config_options << '--disable-system-libraries'
+  e.cross_platform = CROSS_RUBY_PLATFORMS
+  e.cross_compiling do |spec|
+    spec.files.reject! { |path| File.fnmatch?('ports/*', path) }
+    spec.dependencies.reject! { |dep| dep.name == 'mini_portile2' }
+  end
+end
+
 RSpec::Core::RakeTask.new(:spec)
 
-def gem_build_path
-  File.join 'pkg', RE2_GEM_SPEC.full_name
+namespace 'gem' do
+  def gem_builder(platform)
+    # use Task#invoke because the pkg/*gem task is defined at runtime
+    Rake::Task["native:#{platform}"].invoke
+    Rake::Task["pkg/#{RE2_GEM_SPEC.full_name}-#{Gem::Platform.new(platform)}.gem"].invoke
+  end
+
+  CROSS_RUBY_PLATFORMS.each do |platform|
+    # The Linux x86 image (ghcr.io/rake-compiler/rake-compiler-dock-image:1.3.0-mri-x86_64-linux)
+    # is based on CentOS 7 and has two versions of cmake installed:
+    # a 2.8 version in /usr/bin and a 3.25 in /usr/local/bin. The latter is needed by abseil.
+    cmake = platform == 'x86_64-linux' ? '/usr/local/bin/cmake' : 'cmake'
+    desc "build native gem for #{platform} platform"
+    task platform do
+      RakeCompilerDock.sh <<~SCRIPT, platform: platform, verbose: true
+        gem install bundler --no-document &&
+        bundle &&
+        CMAKE=#{cmake} bundle exec rake gem:#{platform}:builder MAKE='nice make -j`nproc`'
+      SCRIPT
+    end
+
+    namespace platform do
+      desc "build native gem for #{platform} platform (guest container)"
+      task 'builder' do
+        gem_builder(platform)
+      end
+    end
+  end
+
+  desc 'build all native gems'
+  task 'native' => CROSS_RUBY_PLATFORMS
 end
 
 def add_file_to_gem(relative_source_path)
@@ -36,6 +87,10 @@ def add_file_to_gem(relative_source_path)
   safe_ln relative_source_path, dest_path
 
   RE2_GEM_SPEC.files << relative_source_path
+end
+
+def gem_build_path
+  File.join 'pkg', RE2_GEM_SPEC.full_name
 end
 
 def add_vendored_libraries
--- a/Rakefile
+++ b/Rakefile
@@ -12,10 +12,44 @@ CLOBBER.include FileList['**/tmp'],
                 FileList['tmp/']
 CLOBBER.add("ports/*").exclude(%r{ports/archives$})
 
+RE2_GEM_SPEC = Gem::Specification.load('re2.gemspec')
+
 Rake::ExtensionTask.new('re2')
+
+Gem::PackageTask.new(RE2_GEM_SPEC) do |p|
+  p.need_zip = false
+  p.need_tar = false
+end
 
 RSpec::Core::RakeTask.new(:spec)
 
+def gem_build_path
+  File.join 'pkg', RE2_GEM_SPEC.full_name
+end
+
+def add_file_to_gem(relative_source_path)
+  dest_path = File.join(gem_build_path, relative_source_path)
+  dest_dir = File.dirname(dest_path)
+
+  mkdir_p dest_dir unless Dir.exist?(dest_dir)
+  rm_f dest_path if File.exist?(dest_path)
+  safe_ln relative_source_path, dest_path
+
+  RE2_GEM_SPEC.files << relative_source_path
+end
+
+def add_vendored_libraries
+  dependencies = YAML.load_file(File.join(File.dirname(__FILE__), 'dependencies.yml'))
+  abseil_archive = File.join('ports', 'archives', "#{dependencies['abseil']['version']}.tar.gz")
+  libre2_archive = File.join('ports', 'archives', "re2-#{dependencies['libre2']['version']}.tar.gz")
+
+  add_file_to_gem(abseil_archive)
+  add_file_to_gem(libre2_archive)
+end
+
+task gem_build_path do
+  add_vendored_libraries
+end
+
 task :spec    => :compile
 task :default => :spec
-
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -114,6 +114,21 @@ def cmake_compile_flags(host)
   ]
 end
 
+# By default, mini_portile2 might add an unnecessary option:
+# https://github.com/flavorjones/mini_portile/blob/5084a2aeab12076f534cf0cabc81a4d5f84b5c25/lib/mini_portile2/mini_portile_cmake.rb#L17
+def delete_cmake_generator_option!(options)
+  indices = []
+
+  options.each_with_index do |element, index|
+    if element == '-G' && index + 1 < options.length
+      indices << index
+      indices << index + 1
+    end
+  end
+
+  indices.reverse_each { |index| options.delete_at(index) }
+end
+
 #
 #  main
 #
@@ -240,6 +255,7 @@ def process_recipe(name, version)
       '-DCMAKE_INSTALL_LIBDIR=lib'
     ]
     recipe.configure_options += cmake_compile_flags(recipe.host)
+    delete_cmake_generator_option!(recipe.configure_options)
 
     yield recipe
 
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -15,12 +15,12 @@ RE2_HELP_MESSAGE = <<~HELP
 
     Flags that are always valid:
 
-      --use-system-libraries
       --enable-system-libraries
-          Use system libraries instead of building and using the packaged libraries. This is the default.
+          Use system libraries instead of building and using the packaged libraries.
 
       --disable-system-libraries
-          Use the packaged libraries, and ignore the system libraries. This overrides `--use-system-libraries`.
+          Use the packaged libraries, and ignore the system libraries. This is the default.
+
 
     Flags only used when using system libraries:
 
@@ -51,9 +51,7 @@ HELP
 #  utility functions
 #
 def config_system_libraries?
-  enable_config("system-libraries", true) do |_, default|
-    arg_config("--use-system-libraries", default)
-  end
+  enable_config("system-libraries", ENV.key?('RE2_USE_SYSTEM_LIBRARIES'))
 end
 
 def concat_flags(*args)
@@ -67,6 +65,15 @@ end
 
 def darwin?
   RbConfig::CONFIG["target_os"].include?("darwin")
+end
+
+def windows?
+  # Use =~ instead of match? to preserve Ruby 2.3 compatibility
+  RbConfig::CONFIG["target_os"] =~ /mingw|mswin/
+end
+
+def freebsd?
+  RbConfig::CONFIG["target_os"].include?("freebsd")
 end
 
 def target_host
@@ -84,13 +91,17 @@ end
 # `--host` parameter.  However, we don't have that feature with
 # cmake. Search for the right compiler for the target architecture using
 # some basic heruistics.
+# See https://github.com/flavorjones/mini_portile/issues/128.
 def find_c_and_cxx_compilers(host)
+  c_compiler = ENV["CC"]
+  cxx_compiler = ENV["CXX"]
+
   if darwin?
-    c_compiler = 'clang'
-    cxx_compiler = 'clang++'
+    c_compiler ||= 'clang'
+    cxx_compiler ||='clang++'
   else
-    c_compiler = 'gcc'
-    cxx_compiler = 'g++'
+    c_compiler ||= 'gcc'
+    cxx_compiler ||= 'g++'
   end
 
   c_platform_compiler = "#{host}-#{c_compiler}"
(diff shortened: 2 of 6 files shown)


stanhu:
@mudge I've tested this on Linux and macOS. If you're happy with this approach, I suggest:

1. Adding CI to build with `bundle exec rake compile -- --disable-system-libraries`.
2. Modifying the `Rakefile` to include the downloaded `.tar.gz` files in the gem (such as https://github.com/kwilczynski/ruby-magic/blob/b781dd6b03f45e2e7c31939f9c034d568992a27e/Rakefile#L126-L137).
3. Consider building precompiled native gems with `rack-compiler`.

re2.gemspec
```diff
@@ -29,4 +29,5 @@ Gem::Specification.new do |s|
   ]
   s.add_development_dependency("rake-compiler", "~> 0.9")
   s.add_development_dependency("rspec", "~> 3.2")
+  s.add_runtime_dependency("mini_portile2", "~> 2.8.2") # keep version in sync with extconf.rb
```
stanhu, line 32: Unfortunately this gem needs Ruby 2..3 to work: https://github.com/mudge/re2/actions/runs/5525303526/jobs/10078757029#step:6:27

mudge:
This is a huge amount of work; thanks for putting in the effort, @stanhu!

Re your checklist:

1. Agreed, it looks like you have the matrix set up but we should be testing all supported Ruby versions against supported system re2 ABI versions as well as all Ruby versions against the vendored dependencies.
2. This was going to be my first question: could we vendor the dependencies directly so users only need to download a single gem?
3. This would be the ultimate goal: users don't need to compile anything at all and just download a binary that Just Works on their system of choice.

With the dependency on MiniPortile meaning we drop support for Ruby < 2.3, I'd consider this worthy of a major version bump (i.e. re2 2.0) and, if we're breaking compatibility, perhaps it would make sense for the vendored libraries to be the default. That better justifies including the sources otherwise it seems a little bloated to include dependencies that will never be used. It also makes the precompiled gems make a little more sense: the default expectation is that you can install the gem without having to install re2 or abseil yourself.

mudge:
> I think that's what happening now in https://github.com/mudge/re2/pull/67/checks. Was there something I missed?
No no, just that we agree on what we should be testing.

(I suppose it is wishful thinking to only require Ruby >= 2.3 if you're opting into the vendored dependencies as we need the gem dependency at install time.)

> I would hesitate making the vendored libraries the default until precompiled gems are available for a number of reasons:
> 
> 1. Building `abseil` takes a minute or two.
> 2. Users may not have a recent enough C++ compiler and/or `cmake`.
Agreed, I was imagining that we'd only make it the default when and only when precompiled gems are available. That way, the common user experience is a fast install but it degrades to a slow compile if a precompiled gem isn't available for your platform.

> > I should note that on ARM64 platforms, precompiled gems built with rake-compiler-dock require glibc 2.29 (see [rake-compiler/rake-compiler-dock#68](https://github.com/rake-compiler/rake-compiler-dock/pull/68)).
I'm unfamiliar with precompiled gems, I don't suppose there's a way to only install one if your platform meets more than an CPU architecture requirement?

> I have a change I'm testing right now that would generate precompiled gems, but then you'll probably want to have some automation to build and push these gems automatically (such as https://github.com/y-crdt/yrb/blob/main/.github/workflows/gem.yml).
Agreed.

> How would you like to proceed? Do you want to tackle everything in this pull request, or get this merged first?
Would it be helpful for me to make a new `v2.0.x` branch and we can start merging PRs into that?

stanhu:
> Would it be helpful for me to make a new v2.0.x branch and we can start merging PRs into that?

Either way.

I'm still ironing out some cross-compiling issues for arm64.

stanhu:
@mudge Ok, I fixed the cross-compilation issues. Turns out using `cmake` is a bit trickier than the standard `configure` and `make`. If you run `bundle exec rake gem:native` now, this should build precompiled gems for four platforms:

-  aarch64-linux
-  arm64-darwin
-  x86_64-darwin
-  x86_64-linux

Missing from this list:

- arm-linux
- x64-mingw-ucrt
- x64-mingw32
- x86-linux
- x86-mingw32
- jruby

I'm not sure what's going on with the Windows builds yet, but it should be possible given `grpc` compiles `abseil` and ships with precompiled Windows gems.

Also, the pre-compiled gems only ship with Ruby 2.7 - 3.2 binaries. Do you want to include 2.3 and up?

ext/re2/extconf.rb
```diff
@@ -6,6 +6,134 @@
...
+      LIBS
+          Add this string to the flags passed to the linker
+HELP
+
+#
+#  utility functions
+#
+def config_system_libraries?
+  enable_config("system-libraries", true) do |_, default|
+    arg_config("--use-system-libraries", default)
```
flavorjones, line 55: In Nokogiri, `--use-system-libraries` only exists for backwards compatibility, I don't think it's necessary to provide more than one commandline flag for this.

ext/re2/extconf.rb
```diff
@@ -6,6 +6,134 @@
...
+
+      LIBS
+          Add this string to the flags passed to the linker
+HELP
+
+#
+#  utility functions
+#
+def config_system_libraries?
+  enable_config("system-libraries", true) do |_, default|
```
flavorjones, line 54: You may want to offer an environment variable to switch this, as well? Especially with bundler intermediating, it's sometimes easier (e.g., in CI job setup) to set an env var than to set up the bundler config. Totally up to you.

ext/re2/extconf.rb
```diff
@@ -6,6 +6,134 @@
...
+  # needed to ensure cross-compilation with CMake targets the right CPU and compilers
+  [
+    "-DCMAKE_SYSTEM_PROCESSOR=#{RbConfig::CONFIG['target_cpu']}",
+    "-DCMAKE_SYSTEM_NAME=#{system_name}",
+    "-DCMAKE_C_COMPILER=#{c_compiler}",
+    "-DCMAKE_CXX_COMPILER=#{cxx_compiler}"
+  ]
+end
+
+# By default, mini_portile2 might add an unnecessary option:
```
flavorjones, line 117: :thinking: Is this something we should fix upstream?

flavorjones:
:wave: @mudge asked me to take a look. I can't see anything that stands out as needing fixing, though I made a few comments.

What I think is missing, though, is more test coverage.

I would recommend **round-trip tests for the precompiled gems**:
- for each architecture:
  - build a precompiled gem (e.g. https://github.com/flavorjones/ruby-c-extensions-explained/blob/main/.github/workflows/precompiled.yml#L161)
  - for each supported ruby version
    - install the gem and run tests against it (preferably the entire suite) (e.g. https://github.com/flavorjones/ruby-c-extensions-explained/blob/main/.github/workflows/precompiled.yml#L191)
- also include `musl` target systems for the linux builds to make sure you won't get bitten by any glibc/musl incompatibilities.

And a similar **round-trip test for the vanilla "ruby" platform gem**, just to ensure something's not been broken, e.g. the gemspec file manifest.

Finally -- and this is optional and extra credit -- it may be worth writing tests to verify
- **verify what's in the gem manifest** -- are there any surprising files that you don't need?
  - example at https://github.com/sparklemotion/nokogiri/blob/main/scripts/test-gem-file-contents
- **verify what symbols are being exported** by the precompiled C extension
  - I strongly suggest checking that one and only one symbol be exported, `Init_re2`
  - you can check this in a test, too: https://github.com/ruby/yarp/blob/main/test/library_symbols_test.rb#L93

I'd be happy to help write these! I have freshly-loaded context from gemifying the YARP parser. Just let me know what you want assistance with.

ext/re2/extconf.rb
stanhu: Thanks for that context!

ext/re2/extconf.rb
stanhu: Done, thanks.

stanhu:
@flavorjones Thanks so much for the review and all your work on mini_portile2!

> What I think is missing, though, is more test coverage.

Very true! I just worked out all the kinks with cross-compiling with `cmake` yesterday.

> I would recommend round-trip tests for the precompiled gems

Thanks for the links! That will help move this faster.

> also include musl target systems for the linux builds to make sure you won't get bitten by any glibc/musl incompatibilities.

I was wondering why Nokogiri didn't ship a musl build. Is that because the existing builds are compatible? I attempted to ship a musl gem in another project, but we found we had to require RubyGems >= v3.3.222 because some of our platforms attempted to install the musl version instead of glibc due to https://github.com/rubygems/rubygems/pull/5852.

> And a similar round-trip test for the vanilla "ruby" platform gem, just to ensure something's not been broken, e.g. the gemspec file manifest.

Would love to help with this. I've spent enough time working on this that I'd appreciate more help.

stanhu:
> I was wondering why Nokogiri didn't ship a musl build. Is that because the existing builds are compatible?

To answer my own question:, it looks like on Alpine 3.17:

```
/ # ldd ./usr/lib/ruby/gems/3.1.0/gems/nokogiri-1.15.3-x86_64-linux/lib/nokogiri/3.1/nokogiri.so | more
        /lib/ld-musl-x86_64.so.1 (0x7f42fbe51000)
        libm.so.6 => /lib/ld-musl-x86_64.so.1 (0x7f42fbe51000)
        libdl.so.2 => /lib/ld-musl-x86_64.so.1 (0x7f42fbe51000)
Error relocating ./usr/lib/ruby/gems/3.1.0/gems/nokogiri-1.15.3-x86_64-linux/lib/nokogiri/3.1/nokogiri.so        libc.so.6 => /lib/ld-musl-x86_64.so.1 (0x7f42fbe51000)
<snip>
```

However, since abseil needs libstdc++ and other libraries, the re2 precompiled gem currently fails miserably:

```shell
/ # ldd /usr/lib/ruby/gems/3.1.0/gems/re2-1.7.0-x86_64-linux/lib/re2.so
        /lib/ld-musl-x86_64.so.1 (0x7f4720899000)
        librt.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f4720899000)
        libruby.so.3.1 => /usr/lib/libruby.so.3.1 (0x7f472026a000)
Error loading shared library libstdc++.so.6: No such file or directory (needed by /usr/lib/ruby/gems/3.1.0/gems/re2-1.7.0-x86_64-linux/lib/re2.so)
Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /usr/lib/ruby/gems/3.1.0/gems/re2-1.7.0-x86_64-linux/lib/re2.so)
Error loading shared library libgcc_s.so.1: No such file or directory (needed by /usr/lib/ruby/gems/3.1.0/gems/re2-1.7.0-x86_64-linux/lib/re2.so)
```

Do we need to ship a precompiled musl extension, or is there a better way?

stanhu:
> Still, this isn't a great experience for Alpine users. Do we need to ship a precompiled musl extension, or is there a better way?

I checked what the `grpc` gem does. The precompiled extension doesn't appear to need `libstdc++`, but it does need the `apk add gcompat` to work on Alpine.

Rakefile
```diff
@@ -1,10 +1,121 @@
...
+
+RE2_GEM_SPEC = Gem::Specification.load('re2.gemspec')
+
+Gem::PackageTask.new(RE2_GEM_SPEC) do |p|
+  p.need_zip = false
+  p.need_tar = false
+end
 
-Rake::ExtensionTask.new('re2')
+CROSS_RUBY_VERSIONS = %w[3.2.0 3.1.0 3.0.0 2.7.0].join(':')
```
stanhu, line 25: Do we really want to support all the way to Ruby 2.3.0? I'm inclined to keep it at 2.7.0 and set the gemspec to:

```ruby
spec.required_ruby_version = ">= 2.7.0", 
```

That way no one can accidentally install a precompiled gem with the wrong Ruby version.

flavorjones:
> I was wondering why Nokogiri didn't ship a musl build. Is that because the existing builds are compatible?

Nokogiri builds are compatible with musl (and we did jump through some hoops to do that), but do require gcompat on some systems. You found the relatively recent feature added that supports non-glibc linux platforms, but I haven't had to play with it yet. Formal musl support has been a low priority compared to some of the other work that needs to be done. :shrug: 

> Do we really want to support all the way to Ruby 2.3.0?

I recommend only precompiling for supported Rubies. Nokogiri and Sqlite3 have adopted a policy of only shipping 4 versions max (3 in current support, plus the one most recently out-of-support).

A precompiled extension is a good "carrot" to encourage people to upgrade Ruby. (Also, my experience has been that a disproportionate number of support issues come from people running old versions of Ruby.)

flavorjones:
> Would love to help with this. I've spent enough time working on this that I'd appreciate more help.

Please point me at what you'd like me to do. I'll have some time over the next few days, and I will try to help as much as I can. Just don't want to duplicate anything you've already done!

stanhu:
> Please point me at what you'd like me to do. I'll have some time over the next few days, and I will try to help as much as I can. Just don't want to duplicate anything you've already done!

@flavorjones I'm going to take a break from this work since I have to focus on other things, but I would love help with all the round-trip tests you mentioned!

flavorjones:
@stanhu would love to help ... you want me to just continue this branch in another PR? or add me as a collaborator on your fork?

stanhu:
@flavorjones I'll just add you to my fork. At some point perhaps @mudge can just merge this to the `v2.0.x` branch and work from there.

Rakefile
mudge: If there is no way for us to only ship precompiled binaries for 2.7+ while still supporting system libraries for Rubies older than that then I'm happy to follow @flavorjones' lead with nokogiri and sqlite3 and set the required Ruby for the whole gem to match.

mudge:
I'll merge this into `v2.0.x` now to make it a bit easier to work in parallel. Please feel free to raise issues for the 2.0 milestone if it'd be easier to keep track of everything that way.

mudge:
I've raised the issues mentioned above and attached them to the 2.0 milestone: https://github.com/mudge/re2/milestone/1

flavorjones:
:+1: Just a note that it may be a few days before I'm able to circle back on the testing issues (I have some other OSS work that is blocking other folks that I need to prioritize), but I will get there.

mudge:
@stanhu I just tried running `bundle exec rake clobber compile` on the `v2.0.x` branch on my 14" MacBook Pro (M1 Max) but got the following error:

```
$ bundle exec rake clobber compile
rm -r pkg
mkdir -p tmp/arm64-darwin22/re2/3.2.2
cd tmp/arm64-darwin22/re2/3.2.2
/Users/mudge/.rubies/ruby-3.2.2/bin/ruby -I. -r.rake-compiler-siteconf.rb ../../../../ext/re2/extconf.rb --disable-system-libraries
Building re2 using packaged libraries.
Using mini_portile version 2.8.2
checking for arm64-apple-darwin22-clang... no
checking for clang... yes
checking for arm64-apple-darwin22-clang++... no
checking for clang++... yes
---------- IMPORTANT NOTICE ----------
Building re2 with a packaged version of abseil-20230125.3.
Configuration options: -DCMAKE_CXX_STANDARD\=17 -DCMAKE_POSITION_INDEPENDENT_CODE\=ON -DCMAKE_INSTALL_LIBDIR\=lib -DCMAKE_SYSTEM_PROCESSOR\=arm64 -DCMAKE_SYSTEM_NAME\=Darwin -DCMAKE_C_COMPILER\=clang -DCMAKE_CXX_COMPILER\=clang++ -DABSL_PROPAGATE_CXX_STD\=ON
Extracting 20230125.3.tar.gz into tmp/arm64-apple-darwin22/ports/abseil/20230125.3... OK
Running 'configure' for abseil 20230125.3... OK
Running 'compile' for abseil 20230125.3... OK
Running 'install' for abseil 20230125.3... OK
Activating abseil 20230125.3 (from /Users/mudge/Projects/re2/ports/arm64-apple-darwin22/abseil/20230125.3)...
Using mini_portile version 2.8.2
checking for arm64-apple-darwin22-clang... no
checking for clang... yes
checking for arm64-apple-darwin22-clang++... no
checking for clang++... yes
---------- IMPORTANT NOTICE ----------
Building re2 with a packaged version of libre2-2023-07-01.
Configuration options: -DCMAKE_CXX_STANDARD\=17 -DCMAKE_POSITION_INDEPENDENT_CODE\=ON -DCMAKE_INSTALL_LIBDIR\=lib -DCMAKE_SYSTEM_PROCESSOR\=arm64 -DCMAKE_SYSTEM_NAME\=Darwin -DCMAKE_C_COMPILER\=clang -DCMAKE_CXX_COMPILER\=clang++ -DCMAKE_PREFIX_PATH\=/Users/mudge/Projects/re2/ports/arm64-apple-darwin22/abseil/20230125.3 -DCMAKE_CXX_FLAGS\=-DNDEBUG
Extracting re2-2023-07-01.tar.gz into tmp/arm64-apple-darwin22/ports/libre2/2023-07-01... OK
Running 'configure' for libre2 2023-07-01... OK
Running 'compile' for libre2 2023-07-01... OK
Running 'install' for libre2 2023-07-01... OK
Activating libre2 2023-07-01 (from /Users/mudge/Projects/re2/ports/arm64-apple-darwin22/libre2/2023-07-01)...
checking for whether -L/Users/mudge/Projects/re2/ports/arm64-apple-darwin22/libre2/2023-07-01/lib is accepted as LDFLAGS... yes
checking for whether -L/Users/mudge/Projects/re2/ports/arm64-apple-darwin22/abseil/20230125.3/lib is accepted as LDFLAGS... yes
checking for whether -pthread is accepted as LDFLAGS... yes
checking for whether -lre2 is accepted as LDFLAGS... yes
checking for whether -labsl_flags is accepted as LDFLAGS... yes
checking for whether -labsl_flags_internal is accepted as LDFLAGS... yes
checking for whether -labsl_flags_marshalling is accepted as LDFLAGS... yes
checking for whether -labsl_flags_reflection is accepted as LDFLAGS... yes
checking for whether -labsl_flags_private_handle_accessor is accepted as LDFLAGS... yes
checking for whether -labsl_flags_commandlineflag is accepted as LDFLAGS... yes
checking for whether -labsl_flags_commandlineflag_internal is accepted as LDFLAGS... yes
checking for whether -labsl_flags_config is accepted as LDFLAGS... yes
checking for whether -labsl_flags_program_name is accepted as LDFLAGS... yes
checking for whether -labsl_cord is accepted as LDFLAGS... yes
checking for whether -labsl_cordz_info is accepted as LDFLAGS... yes
checking for whether -labsl_cord_internal is accepted as LDFLAGS... yes
checking for whether -labsl_cordz_functions is accepted as LDFLAGS... yes
checking for whether -labsl_cordz_handle is accepted as LDFLAGS... yes
checking for whether -labsl_crc_cord_state is accepted as LDFLAGS... yes
checking for whether -labsl_crc32c is accepted as LDFLAGS... yes
checking for whether -labsl_crc_internal is accepted as LDFLAGS... yes
checking for whether -labsl_crc_cpu_detect is accepted as LDFLAGS... yes
checking for whether -labsl_hash is accepted as LDFLAGS... yes
checking for whether -labsl_city is accepted as LDFLAGS... yes
checking for whether -labsl_bad_variant_access is accepted as LDFLAGS... yes
checking for whether -labsl_low_level_hash is accepted as LDFLAGS... yes
checking for whether -labsl_raw_hash_set is accepted as LDFLAGS... yes
checking for whether -labsl_hashtablez_sampler is accepted as LDFLAGS... yes
checking for whether -labsl_exponential_biased is accepted as LDFLAGS... yes
checking for whether -labsl_bad_optional_access is accepted as LDFLAGS... yes
checking for whether -labsl_str_format_internal is accepted as LDFLAGS... yes
checking for whether -labsl_synchronization is accepted as LDFLAGS... yes
checking for whether -labsl_graphcycles_internal is accepted as LDFLAGS... yes
checking for whether -labsl_stacktrace is accepted as LDFLAGS... yes
checking for whether -labsl_symbolize is accepted as LDFLAGS... yes
checking for whether -labsl_debugging_internal is accepted as LDFLAGS... yes
checking for whether -labsl_demangle_internal is accepted as LDFLAGS... yes
checking for whether -labsl_malloc_internal is accepted as LDFLAGS... yes
checking for whether -labsl_time is accepted as LDFLAGS... yes
checking for whether -labsl_civil_time is accepted as LDFLAGS... yes
checking for whether -labsl_strings is accepted as LDFLAGS... yes
checking for whether -labsl_strings_internal is accepted as LDFLAGS... yes
checking for whether -labsl_base is accepted as LDFLAGS... yes
checking for whether -labsl_spinlock_wait is accepted as LDFLAGS... yes
checking for whether -labsl_int128 is accepted as LDFLAGS... yes
checking for whether -labsl_throw_delegate is accepted as LDFLAGS... yes
checking for whether -labsl_raw_logging_internal is accepted as LDFLAGS... yes
checking for whether -labsl_log_severity is accepted as LDFLAGS... yes
checking for whether -labsl_time_zone is accepted as LDFLAGS... yes
checking for -lstdc++... yes
checking for stdint.h... yes
checking for rb_str_sublen()... yes
checking for -lre2... yes
checking for re2 that requires explicit C++ version flag... yes
checking for re2 that compiles with c++20 standard... yes
checking for RE2::Match() with endpos argument... yes
checking for RE2::Set::Match() with error information... yes
creating Makefile
cd -
cd tmp/arm64-darwin22/re2/3.2.2
/usr/bin/make
compiling ../../../../ext/re2/re2.cc
../../../../ext/re2/re2.cc:10:10: fatal error: 're2/re2.h' file not found
#include <re2/re2.h>
         ^~~~~~~~~~~
1 error generated.
make: *** [re2.o] Error 1
rake aborted!
Command failed with status (2): [/usr/bin/make...]
/Users/mudge/.gem/ruby/3.2.2/gems/rake-compiler-1.2.3/lib/rake/extensiontask.rb:201:in `block (2 levels) in define_compile_tasks'
/Users/mudge/.gem/ruby/3.2.2/gems/rake-compiler-1.2.3/lib/rake/extensiontask.rb:200:in `block in define_compile_tasks'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/cli/exec.rb:58:in `load'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/cli/exec.rb:58:in `kernel_load'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/cli/exec.rb:23:in `run'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/cli.rb:492:in `exec'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/cli.rb:34:in `dispatch'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/cli.rb:28:in `start'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/exe/bundle:45:in `block in <top (required)>'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/lib/bundler/friendly_errors.rb:117:in `with_friendly_errors'
/Users/mudge/.gem/ruby/3.2.2/gems/bundler-2.4.12/exe/bundle:33:in `<top (required)>'
/Users/mudge/.gem/ruby/3.2.2/bin/bundle:25:in `load'
/Users/mudge/.gem/ruby/3.2.2/bin/bundle:25:in `<main>'
Tasks: TOP => compile => compile:arm64-darwin22 => compile:re2:arm64-darwin22 => copy:re2:arm64-darwin22:3.2.2 => tmp/arm64-darwin22/re2/3.2.2/re2.bundle
(See full trace by running task with --trace)
```

Am I missing something?

stanhu:
@mudge You probably need the changes in #76, namely aa0bdb616498e02e705e5af6df20a03c520c7872.

stanhu:
I extracted the essential changes in https://github.com/mudge/re2/pull/76 into https://github.com/mudge/re2/pull/77.

PR #80, opened 2023-07-27
Fix Windows gem tests

stanhu:
The script for installing the package did not actually execute because `pwsh` was loaded instead of `bash`.

As of today, the rake-compiler-dock image for:

1. `mri-x64-mingw-ucrt` only supports Ruby 3.1 and 3.2.
2. `mri-x64-mingw32` only supports Ruby 2.4 - 3.0.

### DIFF
--- a/.github/workflows/precompiled.yml
+++ b/.github/workflows/precompiled.yml
@@ -310,6 +310,7 @@ jobs:
           name: cruby-x64-mingw32-gem
           path: gems
       - run: ./scripts/test-gem-install gems
+        shell: bash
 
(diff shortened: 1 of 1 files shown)


stanhu:
Ok, https://github.com/mudge/re2/actions/runs/5677907397/job/15387285822?pr=80 is failing since Ruby 3.1 and 3.2 aren't supported. This matches Nokogiri (https://rubygems.org/gems/nokogiri/versions/1.15.3-x64-mingw32), and explains why the `.cross_rubies` file is used.

mudge (approved):

PR #83, opened 2023-07-28
Drop x86-mingw32 from rake gem:native build

stanhu:
Runing `rake gem:x86-mingw32` currently fails with:

```
linking shared-object re2.so
/usr/bin/i686-w64-mingw32-ld: /tmp/re2/ports/i386-mingw32/i686-w64-mingw32/abseil/20230125.3/lib/libabsl_symbolize.a(symbolize.cc.obj):symbolize.cc:(.text+0x106): undefined reference to `_imp__SymFromAddr@20'
collect2: error: ld returned 1 exit status
```

`SymFromAddr` comes from the Windows Debug Help Library.  It's not clear to me whether the 32-bit MinGW ships with the `dbghelp` library, so drop this for now.


mudge:
Shall we close this in favour of https://github.com/mudge/re2/pull/84?

PR #85, opened 2023-08-11
2.0: Vendor re2 and its dependencies and drop support for older Ruby versions

mudge:
I'm raising this so we can look at all of @stanhu's changes at once and get this merged into `main`, replacing the current version 1 of the gem.

Once we're happy with the changes, I propose we release a [prerelease gem](https://guides.rubygems.org/patterns/#prerelease-gems) (e.g. `2.0.0.beta1`) so we can smoke test with real users on various platforms and Ruby versions before pushing ahead to a full `2.0.0` release.

## To do

- [x] Document native gem dependencies
- [x] Document opting out of the native gem
- [x] Update re2 to [2023-09-01](https://github.com/google/re2/releases/tag/2023-09-01)


.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+      - run: ./scripts/test-gem-install gems
+        shell: bash
+
+  cruby-windows-install-ucrt:
+    needs: ["cruby-package"]
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ["3.1", "3.2"]
+        sys: ["enable", "disable"]
```
mudge, line 131: As above, is `sys` used?

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+          - "x86_64-linux"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: ports/archives
+          key: archives-ubuntu-${{hashFiles('ext/re2/extconf.rb')}}
+      - env:
+          DOCKER_IMAGE: "ghcr.io/rake-compiler/rake-compiler-dock-image:${{needs.rcd_image_version.outputs.rcd_image_version}}-mri-${{matrix.plat}}"
```
mudge, line 168: Rather than doing this dynamically, given we know the version of Rake Compiler Dock we have, could we hard-code it to 1.3.0 instead?

```suggestion
          DOCKER_IMAGE: "ghcr.io/rake-compiler/rake-compiler-dock-image:1.3.0-mri-${{matrix.plat}}"
```

.github/workflows/tests.yml
mudge, line 1: Does this now duplicate tests that are run in `.github/workflows/precompiled.yml` aside from the `--enable-system-libraries` variant? Is it worth combining these two files into a single one or perhaps separating testing from packaging the various versions of the gem?

lib/re2.rb
```diff
@@ -3,5 +3,12 @@
 #
 # Copyright (c) 2010-2014, Paul Mucur (http://mudge.name)
 # Released under the BSD Licence, please see LICENSE.txt
-require "re2.so"
+begin
+  ::RUBY_VERSION =~ /(\d+\.\d+)/
+  require_relative "#{Regexp.last_match(1)}/re2.so"
```
mudge, line 8: I know this comes from Nokogiri but I wonder if it'd be better to avoid the global `Regexp.last_match` (even though it is thread-local):

```suggestion
  require_relative "#{RUBY_VERSION[/(\d+\.\d+)/, 1]}/re2.so"
```

scripts/test-gem-build
```diff
@@ -0,0 +1,32 @@
...
+
+OUTPUT_DIR=$1
+BUILD_NATIVE_GEM=$2
+
+test -e /etc/os-release && cat /etc/os-release
+
+set -x
+
+bundle install --local || bundle install
+bundle exec rake set-version-to-timestamp
```
mudge, line 20: I think I asked this before but why do we need to use a timestamped version rather than, say, using `2.0.0.beta1` or whatever the current version is in `re2.gemspec`? Is it because the gems will be uploaded as artifacts and it might be misleading to people browsing the Actions runs?

scripts/test-gem-build
```diff
@@ -0,0 +1,32 @@
...
+
+if [[ "${BUILD_NATIVE_GEM}" == "ruby" ]] ; then
+  # TODO we're only compiling so that we retrieve tarballs, we can do better.
```
mudge, line 23: Is there a way to download MiniPortile dependencies without also compiling the gem?

scripts/test-gem-install
```diff
@@ -0,0 +1,33 @@
...
+
+if [ -n "${BUNDLE_APP_CONFIG:-}" ] ; then
+  export BUNDLE_CACHE_PATH="${BUNDLE_APP_CONFIG}/cache"
+fi
+
+bundle install --local || bundle install
+
+cd $(dirname `gem which re2`)
+cd ..
+rake spec
```
mudge, line 33: I wonder if we could avoid including all the extra test files in the gem by running the test suite from the checkout but making sure the installed gem is first in the load path rather than the `lib` directory.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+  push:
+    branches:
+      - main
+      - v*.*.x
+    tags:
+      - v*.*.*
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - '*'
```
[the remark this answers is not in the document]
stanhu: Sounds like we can drop it.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
+name: precompiled
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  cancel-in-progress: true
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 8 * * 3" # At 08:00 on Wednesday # https://crontab.guru/#0_8_*_*_3
```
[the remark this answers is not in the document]
stanhu: Yeah, I just pulled that from the template. I think it should be okay to drop this.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+      - v*.*.x
+    tags:
+      - v*.*.*
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - '*'
+
+jobs:
+  precompiled:
```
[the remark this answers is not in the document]
stanhu: Sounds good.

mudge:
@stanhu I’m going to try to make time in the evenings for this so please let me know if there’s anything in particular you want me to look at.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+        with:
+          ruby-version: ${{matrix.ruby}}
+          bundler-cache: true
+      - uses: actions/cache@v3
+        with:
+          path: ports
+          key: ports-${{matrix.runs-on}}-${{hashFiles('ext/re2/extconf.rb')}}
+      - run: bundle exec rake compile spec
+
+  cruby-package:
```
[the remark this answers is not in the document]
stanhu: Yes, exactly. It's also forcing the download of the source tarballs.

.github/workflows/tests.yml
stanhu: > Does this now duplicate tests that are run in .github/workflows/precompiled.yml aside from the --enable-system-libraries variant? 

Looking at https://github.com/mudge/re2/actions/runs/5882973965/job/15954632848, I think this workflow tests against the system version of abseil and some older version of libre2? 

> Is it worth combining these two files into a single one or perhaps separating testing from packaging the various versions of the gem?

Sure, I just kept them separate partly to make it easier to review.

lib/re2.rb
stanhu: That works.

scripts/test-gem-build
stanhu: > Is it because the gems will be uploaded as artifacts and it might be misleading to people browsing the Actions runs?

Yes, and if you're testing locally you avoid the issue of overwriting a properly-installed gem.

scripts/test-gem-build
stanhu: We could implement the download ourselves (such as https://github.com/stanhu/pg_query/blob/fc420181abd2b551d90bb91b240431cf52f85d82/ext/pg_query/extconf.rb#L18-L24) or call `MiniPortile.download` (https://github.com/flavorjones/mini_portile/blob/ee637e0a76e5ae7144949d868948b45d8d858051/lib/mini_portile2/mini_portile.rb#L108C3-L112).

scripts/test-gem-install
stanhu: Yes, that's definitely a possibility.

stanhu:
> @stanhu I’m going to try to make time in the evenings for this so please let me know if there’s anything in particular you want me to look at.

Thanks for the in-depth review! I don't have any specific requests at this point, but if there's anything you need help with please let me know!

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.2"
+          bundler-cache: true
+      - run: ./scripts/test-gem-build gems ruby
+      - uses: actions/upload-artifact@v3
+        with:
+          name: cruby-gem
+          path: gems
+          retention-days: 7
```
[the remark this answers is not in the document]
stanhu: No, we can use the defaults.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+
+  cruby-linux-install:
```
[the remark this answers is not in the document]
stanhu: Yes.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+          name: cruby-gem
+          path: gems
+      - run: ./scripts/test-gem-install gems
+
+  cruby-osx-install:
+    needs: ["cruby-package"]
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ["3.2"]
```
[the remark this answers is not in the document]
stanhu: Sure. The precompiled gems should generally be used on macOS, but it couldn't hurt to add more versions.

.github/workflows/precompiled.yml
```diff
@@ -0,0 +1,350 @@
...
+          path: gems
+      - run: ./scripts/test-gem-install gems
+
+  cruby-windows-install:
+    needs: ["cruby-package"]
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ["2.7", "3.0"]
+        sys: ["enable", "disable"]
```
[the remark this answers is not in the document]
stanhu: No, it can be dropped. It looks like it was originally used to toggle between `--enable-system-libraries` and `--disable-system-libraries`.

.github/workflows/precompiled.yml
stanhu: No, it can be dropped.

.github/workflows/precompiled.yml
stanhu: Yes, that would be fine.

scripts/test-gem-build
mudge: I wonder if we could simplify things by explicitly passing the version during the build. By default, it could be the `ref_name` plus the `run_number` but we could use [an input for `workflow_dispatch`](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) that overrides that so I could, e.g., manually trigger a build for 2.0.0.beta1.

Alternatively, if we didn’t add and rely on `RE2::VERSION`, maybe we could get away with making the gemspec set the version from `ENV.fetch("RE2_VERSION", "2.0.0.beta1)` (or whatever the expected version should be) so we can override it for CI runs but also use it for releases.

scripts/test-gem-build
mudge: Do you think that’d be worth doing as a separate Rake task for the bare, uncompiled Ruby gem to be speed up the build?

scripts/test-gem-build
stanhu: Yeah, that would be a nice improvement.

scripts/test-gem-build
stanhu: Yes. https://github.com/mudge/re2/actions/runs/5882973962/job/15954634773 took 2-3 minutes to download and compile, but we could probably make that finish in less than 30 seconds.

stanhu:
@mudge Do you want help with this pull request? Should I submit improvements to the `v2.0.x` branch for the comments here?

mudge:
@stanhu I can make a start on this tonight and keep you apprised?

mudge:
I’ve consolidated the two workflows into one and tried to make the job names clearer (though the names that show in the GitHub UI still need some work).

stanhu:
Looks good!

mudge:
As well as the outstanding comments above (mostly around the version number and how we run the tests), I think we should update the system dependency tests to use the gem too rather than a checkout. Then the workflow will purely be structured around building various gems and then testing them on as many platforms and Ruby versions as possible.

scripts/test-gem-install
mudge: I think we can use [`actions/checkout`'s `sparse-checkout`](https://github.com/actions/checkout#Fetch-only-the-root-files-and-github-and-src-folder) to do this: only checkout `spec` and `.rspec`, install the gem and appropriate version of RSpec and run `rspec` (avoiding the bundle and other development dependencies only relevant for the build).

Rakefile
```diff
@@ -1,10 +1,145 @@
...
+    # The Linux x86 image (ghcr.io/rake-compiler/rake-compiler-dock-image:1.3.0-mri-x86_64-linux)
+    # is based on CentOS 7 and has two versions of cmake installed:
+    # a 2.8 version in /usr/bin and a 3.25 in /usr/local/bin. The latter is needed by abseil.
+    cmake =
+      case platform
+      when 'x86_64-linux', 'x86-linux'
+        '/usr/local/bin/cmake'
+      else
+        'cmake'
+      end
```
mudge, line 71: Would this work if `CMAKE` is an environment variable? If so, perhaps we could lift this into the workflow only for those images.

stanhu: I wanted to make it possible to build locally with `rake gem:native`, so setting `CMAKE` only in the CI job would prevent that.

scripts/test-gem-build
stanhu: #89 makes it so that this job now finishes in 17 seconds (https://github.com/mudge/re2/actions/runs/6017912200/job/16325121343?pr=89).

stanhu:
> As well as the outstanding comments above (mostly around the version number and how we run the tests), I think we should update the system dependency tests to use the gem too rather than a checkout. Then the workflow will purely be structured around building various gems and then testing them on as many platforms and Ruby versions as possible.

I'm trying this in https://github.com/mudge/re2/pull/90. The downside here is that CI will be a little slower since we have to build the gem, but once https://github.com/mudge/re2/pull/89 is merged that job should be fast.

mudge:
@stanhu if we can update this branch to produce 2.0.0.beta1 gems, I’ll push them to RubyGems and we can get some feedback from users on different platforms.

stanhu:
@mudge Ok, I updated https://github.com/mudge/re2/pull/87.

stanhu:
Hmm, https://github.com/mudge/re2/actions/runs/6114100152/job/16594950941 didn't seem to keep the version number.

mudge:
For now, would it be simpler to forego the timestamping and just produce a gem with whatever is specified in the gemspec?

mudge:
You should be a contributor now so please feel free to update the tag and release once it is producing the correct gem.

stanhu:
Ok, thanks. I believe https://github.com/mudge/re2/pull/91 should fix this.

stanhu:
> For now, would it be simpler to forego the timestamping and just produce a gem with whatever is specified in the gemspec?

Yeah, let's do that. I'll push to this branch then.

stanhu:
@mudge I've updated the releases page with the gems: https://github.com/mudge/re2/releases/tag/v2.0.0.beta1

stanhu:
Could you upload these gems to RubyGems? https://rubygems.org/gems/re2

mudge:
That is now done: https://rubygems.org/gems/re2/versions

I've [announced it on ruby.social](https://ruby.social/@mudge/111029298058288514) but would appreciate you sharing the prerelease so we can test it on as many platforms as possible.

Thank you for your patience in getting this first version published.

mudge:
@stanhu I’ve updated the README but would it also but with having more detailed documentation ala https://github.com/sparklemotion/sqlite3-ruby/blob/master/INSTALLATION.md#native-gems-recommended re the supported platforms and various dependencies there (especially on Windows, etc.)?

mudge:
Re Ruby version support: can you remind me what our constraints are here? MiniPortile2 requires Ruby 2.3 but we’ve opted for 2.7+ because each Ruby version requires us to compile it and include it in the native gem?

If so, could we add 2.6 so that the default Ruby on macOS Monterey gets a native gem too? I noticed when testing on my old 2015 MacBook Pro (that can’t be upgraded beyond Monterey) that the system version was too old for our current beta.

stanhu:
> Re Ruby version support: can you remind me what our constraints are here? MiniPortile2 requires Ruby 2.3 but we’ve opted for 2.7+ because each Ruby version requires us to compile it and include it in the native gem?

Yes, exactly.

> If so, could we add 2.6 so that the default Ruby on macOS Monterey gets a native gem too? I noticed when testing on my old 2015 MacBook Pro (that can’t be upgraded beyond Monterey) that the system version was too old for our current beta.

Yes, we could add 2.6, though it's reached EOL as of April 2022.

stanhu:
Thanks for publishing the gems!

I found one minor issue when compiling the gem from source if a system `libre2` is installed already. It appears that we need to give priority to the `ports` `libre2` directory, just as Nokogiri does in https://github.com/sparklemotion/nokogiri/blob/32c754317c4f3c70395a294ae91aaac58dde0f8f/ext/nokogiri/extconf.rb#L979-L993. I'll push up a fix.

mudge:
> Yes, we could add 2.6, though it's reached EOL as of April 2022.

I don’t mind supporting EOL versions of Ruby as long as there’s usage of it (sadly, it looks like https://stats.rubygems.org/ has stopped updating but the 12m view still shows 2.5 with 26% usage in May).

> I found one minor issue when compiling the gem from source if a system libre2 is installed already.

Good catch: can we add a test for that scenario specifically?

stanhu:
> Good catch: can we add a test for that scenario specifically?

Yes, I'm attempting to reproduce it here: https://github.com/mudge/re2/pull/92

Locally this causes:

```
# bundle exec rake spec
WARNING: rake-compiler found compiled files in 'ext/re2' directory. Please remove them.
/usr/local/bin/ruby -I/usr/local/bundle/gems/rspec-core-3.12.2/lib:/usr/local/bundle/gems/rspec-support-3.12.1/lib /usr/local/bundle/gems/rspec-core-3.12.2/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}

Randomized with seed 30879
/usr/local/bin/ruby: symbol lookup error: /usr/local/bundle/gems/re2-2.0.0.beta1/lib/re2.so: undefined symbol: _ZN3re23RE2C1ESt17basic_string_viewIcSt11char_traitsIcEERKNS0_7OptionsE
```

stanhu:
Ok, reproduced the failure in https://github.com/mudge/re2/actions/runs/6126574194/job/16630866362?pr=92.

I think https://github.com/mudge/re2/pull/93 should fix this.

stanhu:
> I’ve updated the README but would it also be worth having more detailed documentation ala https://github.com/sparklemotion/sqlite3-ruby/blob/master/INSTALLATION.md#native-gems-recommended re the supported platforms and various dependencies there (especially re glibc)?

Yes, definitely. Added f49ccdc9cd454236f6a2fb0c91e8d192ce12187b.

mudge:
@stanhu shall we cut a 2.0.0.beta2 if all is well? Anything else we want to add or change at this point?

stanhu:
@mudge Sounds good to me!

mudge:
I’ll wait for the abseil workaround removal in https://github.com/mudge/re2/pull/96 as I’m looking forward to that before publishing and announcing the new beta.

stanhu:
@mudge I forgot that dropping the workaround requires upgrading abseil as well due to https://github.com/abseil/abseil-cpp/issues/1497. 😄

mudge:
Darn! Oh well, let’s cut the next beta from this branch as-is then.

stanhu:
@mudge Ok, do you want to handle it? I have to go now. https://github.com/mudge/re2/actions/runs/6137906858 has the gems.

mudge:
New beta2 gem now pushed to RubyGems.

stanhu:
> New beta2 gem now pushed to RubyGems.

Awesome, thanks! I have confirmed that all GitLab package builds, including arm64 and Raspberry Pi 32-bit builds from source, were successful with v2.0.0.beta2 (https://gitlab.com/gitlab-org/build/omnibus-gitlab-mirror/-/pipelines/998940657).

mudge:
Is there anything else we'd like to get into this release, e.g. the extra tests from https://github.com/mudge/re2/milestone/1?

Otherwise, let's wait for any bug reports this week and get this merged?

stanhu:
> Is there anything else we'd like to get into this release

No, personally I'm happy where we are now. Thanks!

stanhu:
I think this is ready to go! GitLab.com has been running with `2.0.0.beta2` in production for almost 24 hours with no issues.

mudge:
That’s great news, I’ll see about getting a release out and publicised tomorrow.

stanhu:
In addition, I haven't heard any issues from any developers yet. Whenever Homebrew releases a new `re2` update, I usually see a [number of issues](https://gitlab.com/gitlab-org/gitlab-development-kit/-/issues/1689) that usually resolves with `bundle pristine re2`]. This is especially painful for UX designers who don't know anything about Ruby.

@mudge Thanks for creating this gem and being so responsive!

mudge:
The usability for most users is so much better with the native gems but figuring out exactly how to vendor re2, let alone precompile it for multiple platforms always seemed an insurmountable task to me.

As hopefully everyone can see from this diff, this has been a Herculean contribution from you and vastly improves the gem so thank you very much for your perseverance!

mudge:
Do you know how long it typically takes for users to upgrade their instances of GitLab? How long should we wait for any issues to come out of the woodwork before putting out a proper 2.0 release? I can see around 45,000 downloads (almost all x86_64-linux) from RubyGems so far.

stanhu:
> Do you know how long it typically takes for users to upgrade their instances of GitLab? How long should we wait for any issues to come out of the woodwork before putting out a proper 2.0 release? I can see around 45,000 downloads (almost all x86_64-linux) from RubyGems so far.

The next GitLab release will be on September 22, but I would just go ahead with a proper 2.0 release now since fundamentally little has changed in terms of functionality.

There are various ports such as FreeBSD that install from source, but @mfechner tested building `v2.0.0beta2` with `--enable-system-libraries` and said it built fine: https://pkg.fechner.net/data/132amd64-gitlab/2023-09-13_14h19m36s/logs/rubygem-re2-2.0.0.beta2.log

PR #87, opened 2023-08-15
Bump version to v2.0.0.beta1

stanhu:
Only call `rake set-version-to-timestamp` if we're on a branch. This will make it possible to build precompiled gems with the right version tag.

### DIFF
--- a/scripts/test-gem-build
+++ b/scripts/test-gem-build
@@ -11,13 +11,17 @@ set -e -u
 
 OUTPUT_DIR=$1
 BUILD_NATIVE_GEM=$2
+REF_TYPE=$3
 
 test -e /etc/os-release && cat /etc/os-release
 
 set -x
 
 bundle install --local || bundle install
-bundle exec rake set-version-to-timestamp
+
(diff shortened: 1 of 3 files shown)


scripts/test-gem-build
```diff
@@ -11,13 +11,17 @@ set -e -u
...
+REF_TYPE=$3
 
 test -e /etc/os-release && cat /etc/os-release
 
 set -x
 
 bundle install --local || bundle install
-bundle exec rake set-version-to-timestamp
+
+if [[ "${REF_TYPE}}" != "tag" ]]; then
```
mudge, line 22: Is there an extra closing brace here?

```suggestion
if [[ "${REF_TYPE}" != "tag" ]]; then
```

stanhu: Oops, thanks.

mudge:
Will merging this into 2.0.x produce a 2.0.0.beta1 or will I need to tag it too?

stanhu:
> Will merging this into 2.0.x produce a 2.0.0.beta1 or will I need to tag it too?

Oh, right, yes, you will need to tag it.

PR #88, opened 2023-08-16

stanhu:
The linker order problem in https://github.com/abseil/abseil-cpp/issues/1497 was shipped with abseil-cpp 20230802.0, so upgrade to that.

pkgconf v2.0 fixed https://github.com/pkgconf/pkgconf/issues/268, so we should no longer need to add the abseil linker flags manually.


stanhu:
It seems we can't do this yet because the `pkgconf` version on the build machines are still broken. Will close for now.

PR #89, opened 2023-08-29
Make `rake gem` download source tarballs automatically

stanhu:
Previously `scripts/test-gem-build` would call `rake compile` just to make mini_portile2 download the source tarballs. However, this needlessly compiles abseil and libre2 before building the gem.

To improve this, split the loading of the recipes into a separate file (`ext/re2/recipes.rb`). The `rake prepare` step invokes the `download` method to download the required tarballs in preparation for building the gem.

Now we add a `rake prepare` task that will run if `rake gem` is run.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,6 +5,8 @@ require 'rspec/core/rake_task'
 require 'rake_compiler_dock'
 require 'yaml'
 
+require_relative 'ext/re2/recipes'
+
 CLEAN.include FileList['**/*{.o,.so,.dylib,.bundle}'],
               FileList['**/extconf.h'],
               FileList['**/Makefile'],
@@ -17,6 +19,14 @@ CLOBBER.include FileList['**/tmp'],
 CLOBBER.add("ports/*").exclude(%r{ports/archives$})
 
 RE2_GEM_SPEC = Gem::Specification.load('re2.gemspec')
+
+task :prepare do
+  puts "Preparing project for gem building..."
+  recipes = load_recipes
+  recipes.each { |recipe| recipe.download }
+end
+
+task gem: :prepare
 
 Gem::PackageTask.new(RE2_GEM_SPEC) do |p|
   p.need_zip = false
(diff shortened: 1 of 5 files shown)


ext/re2/recipes.rb
```diff
@@ -0,0 +1,43 @@
+PACKAGE_ROOT_DIR = File.expand_path(File.join(File.dirname(__FILE__), '..', '..'))
```
mudge, line 1:
```suggestion
PACKAGE_ROOT_DIR = File.expand_path('../..', __dir__)
```

Rakefile
```diff
@@ -5,6 +5,8 @@ require 'rspec/core/rake_task'
 require 'rake_compiler_dock'
 require 'yaml'
 
+require_relative 'ext/re2/recipes.rb'
```
mudge, line 8: Is the `.rb` significant here or could we simplify?

```suggestion
require_relative 'ext/re2/recipes'
```

Rakefile
```diff
@@ -18,6 +20,14 @@ CLOBBER.add("ports/*").exclude(%r{ports/archives$})
 
 RE2_GEM_SPEC = Gem::Specification.load('re2.gemspec')
 
+task :prepare do
+  puts "Preparing project for gem building..."
+  recipes = load_recipes
+  recipes.each { |recipe| recipe.download }
```
mudge, line 26: Bit of code golf but...

```suggestion
  load_recipes.each(&:download)
```

mudge (commented):
Big picture question: could we simplify any of the new build code if we vendored the re2 and abseil archives into `ports/archives` and committed them to source control? That way, they'd always be present in CI and we could even include them in the gemspec (unless we were compiling a native gem in which case they are already removed)?

stanhu:
> Big picture question: could we simplify any of the new build code if we vendored the re2 and abseil archives into `ports/archives` and committed them to source control? That way, they'd always be present in CI and we could even include them in the gemspec (unless we were compiling a native gem in which case they are already removed)?

Probably, but do you really want to commit tarballs into Git?

mudge (approved):

PR #90, opened 2023-08-30
Update the system dependency tests to use the built gem

stanhu:
This would better reflect how users might install the gem.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,6 +14,7 @@ on:
 
 jobs:
   compile-and-test-system-dependencies:
+    needs: ["build-cruby-gem"]
     name: System Dependencies - Ruby ${{ matrix.ruby }} - libre2 ABI version ${{ matrix.libre2.soname }}
     runs-on: ubuntu-20.04
     strategy:
@@ -57,8 +58,11 @@ jobs:
         with:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
-      - run: bundle exec rake compile -- --enable-system-libraries
-      - run: bundle exec rake spec
+      - uses: actions/download-artifact@v3
+        with:
+          name: cruby-gem
+          path: gems
+      - run: ./scripts/test-gem-install gems --enable-system-libraries
 
   compile-and-test-vendored-dependencies:
     name: Vendored Dependencies - Ruby ${{ matrix.ruby }} - ${{ matrix.runs-on }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -57,8 +58,11 @@ jobs:
         with:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
```
stanhu, line 60: We could drop the `actions/checkout@v3` line on 44, but then we'd have to set this to `false`. Then we would not having caching available in `scripts/test-gem-install`, but maybe that's okay?

mudge (commented):
Is this a good opportunity to see if we can test the installed gems against a checkout of our tests (and all the Bundler/RubyGems shenanigans that may entail)?

I briefly looked at this and Bundler got in the way as I wanted to use the "system" re2 gem I'd installed and running `bundle exec rspec` would fail as it wanted to install its own version. In the end, running a vanilla `rspec` worked because we control the gems installed in CI but perhaps there is a better way.

One thing I'm keen to do with this release is keep the change to the packaged gem as small as possible: the dream being that the contents of the downloaded gem barely changes for users except the inclusion of `re2.so` in the compiled version or the archives required for MiniPortile in the pure Ruby version. Perhaps a good way to measure this is to look at the amount of change to the gemspec's `files` and see what we can do to minimise that?

On that topic, I wonder how much of the test scripts in `scripts` can be inlined into the GitHub Actions workflow itself (though sharing YAML via anchors isn't always pleasant, we also have the option of [creating our own parameterised actions](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#adding-an-action-from-the-same-repository)).

stanhu:
> One thing I'm keen to do with this release is keep the change to the packaged gem as small as possible: the dream being that the contents of the downloaded gem barely changes for users except the inclusion of re2.so in the compiled version or the archives required for MiniPortile in the pure Ruby version. Perhaps a good way to measure this is to look at the amount of change to the gemspec's files and see what we can do to minimise that?

That's a good goal, but the size of the shared libraries for multiple Ruby versions dwarfs the size of the specs so I'm not sure if this is worth the extra effort. For example, in the x86_64 gem, 97% of the uncompressed size is from the `lib` directory:

```
 % du -h .
 72K	./ext/re2
 72K	./ext
 48K	./spec/re2
 60K	./spec
1.6M	./lib/3.2
1.5M	./lib/2.7
 12K	./lib/re2
1.6M	./lib/3.0
1.6M	./lib/3.1
6.2M	./lib
6.4M	.
```

```mermaid
pie title Uncompressed kilobytes of x86_64 gem
    "lib": 6200
    "ext" : 72
    "spec": 60
    "other": 36
```

mudge (approved):

PR #91, opened 2023-09-07
Fix typo in scripts/test-gem-build

stanhu:
This was causing the timestamp to be used in the tagged version.

### DIFF
--- a/lib/re2/version.rb
+++ b/lib/re2/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module RE2
-  VERSION = "2.0.0.beta1"
+  VERSION = "2.0.0.beta2"
 end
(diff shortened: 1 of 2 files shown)


mudge:
Do we still want this or can we remove the timestamping altogether? There’s a risk of confusion but only for anyone working on the gem.

stanhu:
Either way. I found it a little useful especially when downloading the gems from the artifacts. If they have the timestamps, then there's less risk of pushing gems that are built with the wrong version.

mudge:
Going to close this for now in favour of https://github.com/mudge/re2/pull/85/commits/ec01921829291525927bcaef6b3f4e9f2fd1944b (any excuse to reduce the size of the `v2.0.x` diff at increased risk to me).

PR #92, opened 2023-09-08
ci: Ensure vendored dependencies can co-exist with system libs

stanhu:
Ruby's mkmf appears to list `RbConfig::CONFIG['exec_prefix']/lib` first in the search path when building a C extension. If you have libre2 installed in that path, the linker will use that library instead of the vendored library. This adds a test to ensure this case is covered.


stanhu:
Closing in favor of #93!

PR #93, opened 2023-09-08
Fix static linking when system libraries installed

stanhu:
Ruby's mkmf appears to list `RbConfig::CONFIG['exec_prefix']/lib` first in the search path when building a C extension. If you have libre2 installed in that path, the linker will use that library instead of the vendored library. This causes an undefined symbol error when using the extension:

```shell
/usr/local/bin/ruby: symbol lookup error: /usr/local/bundle/gems/re2-2.0.0.beta1/lib/re2.so: undefined symbol: _ZN3re23RE2C1ESt17basic_string_viewIcSt11char_traitsIcEERKNS0_7OptionsE
```

This commit does two things to ensure the final shared library is compiled statically with the vendored libraries:

1. For `-L<path>` flags, ensure that any `ports` paths are prioritized just in case there are installed libraries that might take precedence.

2. For `-l<lib>` flags, convert the library to the static library with a full path and substitute the absolute static library.  For example, -lre2 maps to `/path/to/ports/<arch>/libre2/<version>/lib/libre2.a`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,10 +15,11 @@ on:
 jobs:
   compile-and-test-system-dependencies:
     needs: ["build-cruby-gem"]
-    name: System Dependencies - Ruby ${{ matrix.ruby }} - libre2 ABI version ${{ matrix.libre2.soname }}
+    name: System Dependencies - ${{ matrix.sys }} vendored libs - Ruby ${{ matrix.ruby }} - libre2 ABI version ${{ matrix.libre2.soname }}
     runs-on: ubuntu-20.04
     strategy:
       matrix:
+        sys: ["enable", "disable"]
         ruby:
           - '3.2'
           - '3.1'
@@ -62,7 +63,9 @@ jobs:
         with:
           name: cruby-gem
           path: gems
-      - run: ./scripts/test-gem-install gems --enable-system-libraries
+      - name: Add a symlink for libre2
+        run: ln -s /usr/lib/libre2.so `ruby -e "puts RbConfig::CONFIG['exec_prefix']"`/lib/libre2.so
+      - run: ./scripts/test-gem-install gems --${{matrix.sys}}-system-libraries
 
   compile-and-test-vendored-dependencies:
     name: Vendored Dependencies - Ruby ${{ matrix.ruby }} - ${{ matrix.runs-on }}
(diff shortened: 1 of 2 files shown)


mudge:
Is any part of this worth upstreaming into MiniPortile2 or is it specific to our setup?

stanhu:
> Is any part of this worth upstreaming into MiniPortile2 or is it specific to our setup?

Good question. I think this is outside the purview of MiniPortile2; Nokogiri does something similar. I suspect `mkmf` could be improved to make it easier to statically link external libraries, but I suspect it's not that common to do this.

ext/re2/extconf.rb
```diff
@@ -128,7 +128,7 @@ def build_extension
   have_header("stdint.h")
   have_func("rb_str_sublen")
 
-  unless have_library("re2")
+  if !static_p and !have_library("re2")
```
mudge, line 131: Are we basically using our own variant of `have_library` when using MiniPortile2 here because we need to override the linker config to give our vendored dependencies priority?

stanhu: Yes, we're linking the `.a` explicitly instead of using `-lre2`. The latter will pick up the system dependency if it's installed in a certain location, and I don't believe there's anything we can do about that.

mudge:
Does this avoid the problems you ran into with https://github.com/mudge/re2/pull/64 because we control the compilation of libre2 and can ensure it generates position-independent code?

stanhu:
> Does this avoid the problems you ran into with https://github.com/mudge/re2/pull/64 because we control the compilation of libre2 and can ensure it generates position-independent code or is there a scenario where we might run into that again?

Yes. We set `cmake` with `-DCMAKE_POSITION_INDEPENDENT_CODE=ON'` in https://github.com/mudge/re2/blob/279d61c0a6af823efb354ebb118c95e75b4bfdb8/ext/re2/recipes.rb#L15.

.github/workflows/tests.yml
```diff
@@ -62,7 +63,9 @@ jobs:
         with:
           name: cruby-gem
           path: gems
-      - run: ./scripts/test-gem-install gems --enable-system-libraries
+      - name: Add a symlink for libre2
+        run: ln -s /usr/lib/libre2.so `ruby -e "puts RbConfig::CONFIG['exec_prefix']"`/lib/libre2.so
```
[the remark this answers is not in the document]
stanhu: I haven't looked, but according to https://github.com/mudge/re2/actions/runs/6126602337/job/16630989217?pr=93, I suspect it's something like `/opt/hostedtoolcache/Ruby/3.1.4/x64`. https://github.com/ruby/setup-ruby/issues/140 describes another way of retrieving the prefix.

PR #95, opened 2023-09-09
Upgrade abseil to 20230802.1

mudge:
RE2 2023-09-01 was released a week ago and there is a more recent LTS version of Abseil for us to use. However, when I initially pushed this to our v2.0.x branch, it looks like there are quite a few build failures to investigate.


mudge:
If it’s a distraction right now, we could skip the abseil upgrade but it’d be good to know why it doesn’t work when the January release does.

stanhu:
It looks like there are two main issues:

1. macOS (https://github.com/mudge/re2/actions/runs/6130653443/job/16639942206): `absel::optional` is aliased to C++17 `std::optional`, but the call to `value()` (introduced in abseil-cpp via https://github.com/abseil/abseil-cpp/commit/751ade00ee347abef5dac7248db851e3f2012e14) is only available starting in macOS 10.14. Even though the macOS SDK 11.1 is used, I'd have to presume the compiler tries to target a lower minimum target version:

```
 [ 97%] Building CXX object absl/status/CMakeFiles/status.dir/status.cc.o
/tmp/d20230909-93-us36r9/tmp/arm64-apple-darwin/ports/abseil/20230802.0/abseil-cpp-20230802.0/absl/status/status.cc:126:51: error: 'value' is unavailable: introduced in macOS 10.14
  if (index.has_value()) return (*payloads)[index.value()].payload;
                                                  ^
```

I think we might be able to set `ABSL_OPTION_USE_STD_OPTIONAL` to 0 to disable this aliasing, or consider raising the macOS minimum version. `clang --help` shows this option:

```
-mmacosx-version-min=<value>
                      Set Mac OS X deployment target
```

2. Windows (https://github.com/mudge/re2/actions/runs/6130653443/job/16639942231):

```
[ 37%] Building CXX object absl/synchronization/CMakeFiles/synchronization.dir/internal/create_thread_identity.cc.obj
In file included from /tmp/d20230909-93-l6kcab/tmp/x86_64-w64-mingw32/ports/abseil/20230802.0/abseil-cpp-20230802.0/absl/synchronization/internal/create_thread_identity.cc:21:
/tmp/d20230909-93-l6kcab/tmp/x86_64-w64-mingw32/ports/abseil/20230802.0/abseil-cpp-20230802.0/absl/synchronization/internal/waiter.h:44:2: error: #error ABSL_WAITER_MODE is undefined
   44 | #error ABSL_WAITER_MODE is undefined
      |  ^~~~~
/tmp/d20230909-93-l6kcab/tmp/x86_64-w64-mingw32/ports/abseil/20230802.0/abseil-cpp-20230802.0/absl/synchronization/internal/waiter.h:51:5: warning: "ABSL_WAITER_MODE" is not defined, evaluates to 0 [-Wundef]
   51 | #if ABSL_WAITER_MODE == ABSL_WAITER_MODE_FUTEX
      |     ^~~~~~~~~~~~~~~~
In file included from /tmp/d20230909-93-l6kcab/tmp/x86_64-w64-mingw32/ports/abseil/20230802.0/abseil-cpp-20230802.0/absl/synchronization/internal/create_thread_identity.cc:21:
/tmp/d20230909-93-l6kcab/tmp/x86_64-w64-mingw32/ports/abseil/20230802.0/abseil-cpp-20230802.0/absl/synchronization/internal/waiter.h:52:16: error: 'FutexWaiter' does not name a type
   52 | using Waiter = FutexWaiter;
      |                ^~~~~~~~~~~
make[2]: *** [absl/synchronization/CMakeFiles/synchronization.dir/build.make:92: absl/synchronization/CMakeFiles/synchronization.dir/internal/create_thread_identity.cc.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:4540: absl/synchronization/CMakeFiles/synchronization.dir/all] Error 2
```

There appears to be a workaround: https://github.com/abseil/abseil-cpp/issues/1510#issuecomment-1694249704.

scripts/test-gem-build
```diff
@@ -23,6 +23,7 @@ if [[ "${BUILD_NATIVE_GEM}" == "ruby" ]] ; then
   bundle exec rake clean
   bundle exec rake gem
 else
+  export MACOSX_DEPLOYMENT_TARGET=10.14
```
stanhu, line 26: I tried hard to avoid this duplication in the `Rakefile`. I originally tried to pass `cmake` `-DCMAKE_CXX_FLAGS=-mmacosx-version-min=10.14` (https://github.com/tpoechtrager/osxcross#deployment-target), but the compiler flags had redundant values, with the latter one overriding the one I set:

```
-mmacosx-version-min=10.14 -mmacosx-version-min=10.13 ...
```

ext/re2/extconf.rb
```diff
@@ -295,17 +295,18 @@ def build_with_system_libraries
...
   -labsl_bad_variant_access
   -labsl_low_level_hash
-  -labsl_raw_hash_set
   -labsl_hashtablez_sampler
   -labsl_exponential_biased
   -labsl_bad_optional_access
   -labsl_str_format_internal
   -labsl_synchronization
   -labsl_graphcycles_internal
+  -labsl_kernel_timeout_internal
```
stanhu, line 309: I was hoping we could drop this workaround in https://github.com/mudge/re2/pull/96, but the runners are still using `pkgconf` 1.9.3. 2.0.0 will fix this: https://github.com/pkgconf/pkgconf/issues/268

stanhu:
@mudge I think this is ready to go.

scripts/test-gem-build
mudge: Does this mean the native gems for macOS will no longer run on High Sierra (10.13) and instead require Mojave (10.14)? If so, perhaps we should add that to the README along with the glibc requirements.

stanhu: Yes. The alternative is to file an upstream fix to abseil to be able control the behavior of `ABSL_OPTION_USE_STD_OPTIONAL` independently of `MACOSX_DEPLOYMENT_TARGET`, but I suspect it's not worth the effort.

mudge:
Updated for the recently released 20230802.1 (note I removed https://github.com/mudge/re2/pull/95/commits/c284a7e2bc9c873ce756b4b39e17568d40f8b6ea just to see if that regression has been fixed but we may need to restore it).

mudge:
@stanhu any chance you could take a look at this? The Windows failures seem different to with patch level 0.

```
undefined reference to `absl::lts_20230802::synchronization_internal::Win32Waiter::Win32Waiter()
```

mudge:
Out of interest, is MacPorts providing a version of Abseil that works with versions of macOS older than 10.14? See https://github.com/judaew/macports-ports/blob/master/devel/abseil/Portfile

stanhu:
`std::optional` was added in C++17, and I see that they have set C++14 in https://github.com/judaew/macports-ports/blob/c536e357838df685f32fdedc7b74dd5d4f0325ce/devel/abseil/Portfile#L68. This might avoid the need to upgrade macOS. Maybe for macOS we can use C++14.

mudge:
> `std::optional` was added in C++17, and I see that they have set C++14 in https://github.com/judaew/macports-ports/blob/c536e357838df685f32fdedc7b74dd5d4f0325ce/devel/abseil/Portfile#L68. This might avoid the need to upgrade macOS. Maybe for macOS we can use C++14.

Good spot. Given Abseil only needs C++14 (and RE2 follows suit), could we get away with removing references to any higher C++ Standard?

stanhu:
Did this branch drop 0b57520e4f8b689bb6212f7d20f73d154a26c057?

mudge:
Yes, I've restored it now (I was trying to see what was and wasn't required with patch level 1).

mudge:
The build is now green and (aside from the necessary workaround for https://github.com/abseil/abseil-cpp/issues/1510) I'm happier with the diff especially as we're not dropping support for any existing users.

mudge:
If you're happy, I'd be tempted to merge this but not include it in a release until RE2 2023-10-01 is out (assuming it has a monthly release cadence now).

mudge:
I've raised https://github.com/mudge/re2/pull/111 to partially automate this sort of thing in future: it's set to run on the first of the month and raise a PR with an updated `dependencies.yml` (if there are any releases). I'm tempted to merge that, trigger it manually and then cherry pick these commits onto the PR that is automatically raised.

stanhu:
> If you're happy, I'd be tempted to merge this but not include it in a release until RE2 2023-10-01 is out (assuming it has a monthly release cadence now).

Sounds good!

https://github.com/mudge/re2/pull/112 didn't run CI. Does the user need to be approved?

mudge:
I think it’s due to https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs I’ll try one of the workarounds and try triggering it again to make sure it runs CI too.

mudge:
Closing in favour of https://github.com/mudge/re2/pull/113 as that PR was raised automatically.

PR #96, opened 2023-09-10

stanhu:


stanhu:
pkgconf doesn't appear to have been updated yet. Closing.

mudge:
Is our only option to wait for GitHub to update the Windows runners?

stanhu:
> Is our only option to wait for GitHub to update the Windows runners? 

We might be able to update `pkgconf` manually.

mudge:
Would it be best to fold this into https://github.com/mudge/re2/pull/95 if that’s a requirement?

stanhu:
> Would it be best to fold this into https://github.com/mudge/re2/pull/95 if that’s a requirement?

Yes. I'm still seeing some failed builds due to `pkgconf`. Filed https://github.com/pkgconf/pkgconf/issues/322.

Issue #74, opened 2023-07-15
Codify encoding behaviour


mudge:
While working on https://github.com/mudge/re2/issues/59 last year and trying to plug gaps in test coverage, I realised the behaviour of string encoding when using Latin and UTF-8 re2 patterns is not especially well-documented. This led me to doubt the correctness of its behaviour but, regardless of that, the current behaviour should be better tested and documented.

Issue #73, opened 2023-07-15
Be liberal in accepting string-likes


mudge:
When working on https://github.com/mudge/re2/issues/59 last year, I noticed that the Set interface raises an error if given non-string arguments even if those argument could happily be coerced to a String. We should consistently use StringValue and StringValuePtr instead.

PR #101, opened 2023-09-15
Increase test coverage, coerce inputs to string and fix encoding inconsistencies

mudge:
Where possible, coerce inputs to strings with StringValue rather than raising a TypeError. This particularly impacts the relatively recent RE2::Set API which was excessively strict about its arguments.

Add test coverage to all parts of the API, better covering edge cases including how encoding is handled based on the encoding of the RE2 pattern.

### DIFF
--- a/ext/re2/re2.cc
+++ b/ext/re2/re2.cc
@@ -648,6 +648,10 @@ static VALUE re2_matchdata_to_s(VALUE self) {
 /*
  * Returns a printable version of the match.
  *
+ * Note RE2 only supports UTF-8 and ISO-8859-1 encoding so strings will be
+ * returned in UTF-8 by default or ISO-8859-1 if the :utf8 option for the
+ * RE2::Regexp is set to false (any other encoding's behaviour is undefined).
+ *
  * @return [String] a printable version of the match
  * @example
  *   m = RE2::Regexp.new('(\d+)').match("bob 123")
@@ -876,6 +880,10 @@ static VALUE re2_regexp_initialize(int argc, VALUE *argv, VALUE self) {
 /*
  * Returns a printable version of the regular expression +re2+.
  *
+ * Note RE2 only supports UTF-8 and ISO-8859-1 encoding so strings will be
+ * returned in UTF-8 by default or ISO-8859-1 if the :utf8 option for the
+ * RE2::Regexp is set to false (any other encoding's behaviour is undefined).
+ *
  * @return [String] a printable version of the regular expression
  * @example
  *   re2 = RE2::Regexp.new("woo?")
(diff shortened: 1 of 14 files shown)


spec/re2_spec.rb
```diff
@@ -15,29 +15,68 @@
...
+    end
+
+    it "returns ISO-8859-1 strings if the pattern is not UTF-8" do
+      original = "Foo".encode("Shift_JIS")
+      replacement = RE2.Replace(original, RE2("oo", :utf8 => false), "ah")
+
+      expect(replacement.encoding).to eq(Encoding::ISO_8859_1)
+    end
+
+    it "returns strings in the encoding of the given String pattern" do
```
mudge, line 60: It’s worth noting this probably isn’t the right behaviour: the pattern will be compiled into an `RE2` but the input will be interpreted as UTF-8 by default so any result will also be UTF-8. (Not to mention this will probably result in undefined behaviour if passing non-UTF-8 data.)

spec/re2_spec.rb
```diff
@@ -53,23 +92,78 @@
...
+    end
+
+    it "returns ISO-8859-1 strings if the pattern is not UTF-8" do
+      original = "Foo".encode("Shift_JIS")
+      replacement = RE2.GlobalReplace(original, RE2("oo", :utf8 => false), "ah")
+
+      expect(replacement.encoding).to eq(Encoding::ISO_8859_1)
+    end
+
+    it "returns strings in the encoding of the given String pattern" do
```
mudge, line 137: As above, this should probably be UTF-8 as that is what RE2 will use.

mudge (commented):
This will close https://github.com/mudge/re2/issues/74 and https://github.com/mudge/re2/issues/73.

spec/re2/scanner_spec.rb
```diff
@@ -21,9 +21,10 @@
   end
 
   describe "#scan" do
-    it "returns the next array of matches" do
+    it "returns the next array of matches", :aggregate_failures do
       r = RE2::Regexp.new('(\w+)')
       scanner = r.scan("It is a truth universally acknowledged")
+
       expect(scanner.scan).to eq(["It"])
```
mudge, line 28: We should also assert the encoding behaviour here.

ext/re2/re2.cc
```diff
@@ -1418,8 +1410,8 @@ static VALUE re2_Replace(VALUE self, VALUE str, VALUE pattern,
     RE2::Replace(&str_as_string, StringValuePtr(pattern),
         StringValuePtr(rewrite));
 
-    return ENCODED_STR_NEW2(str_as_string.data(), str_as_string.size(),
-        pattern);
+    return ENCODED_STR_NEW(str_as_string.data(), str_as_string.size(),
```
mudge, line 1413: Given this change, we should document how RE2 only supports UTF-8 and ISO-8859-1 and anything else is undefined.

ext/re2/re2.cc
```diff
@@ -1418,15 +1461,19 @@ static VALUE re2_Replace(VALUE self, VALUE str, VALUE pattern,
...
+        "UTF-8");
```
mudge, line 1465: @stanhu will this (and the change to `GlobalReplace` below) affect GitLab or are you only using UTF-8 strings with re2?

stanhu: As far as I can tell, GitLab is only using UTF-8 strings with re2.

PR #103, opened 2023-09-20
Tidy up GitHub Actions

mudge:
Try to speed up builds by reducing duplication (no longer testing the Rake extension task but only testing the gems).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,99 +7,14 @@ on:
   push:
     branches:
       - main
-      - v*.*.x
+      - v*.*
     tags:
       - v*.*.*
   pull_request:
 
 jobs:
-  compile-and-test-system-dependencies:
-    needs: ["build-cruby-gem"]
-    name: System Dependencies - ${{ matrix.sys }} vendored libs - Ruby ${{ matrix.ruby }} - libre2 ABI version ${{ matrix.libre2.soname }}
-    runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        sys: ["enable", "disable"]
-        ruby:
-          - '3.2'
-          - '3.1'
-          - '3.0'
-          - '2.7'
-          - '2.6'
-        libre2:
-          - version: "20150501"
-            soname: 0
-          - version: "20200302"
-            soname: 1
-          - version: "20200303"
-            soname: 6
-          - version: "20200501"
-            soname: 7
-          - version: "20200706"
-            soname: 8
-          - version: "20201101"
-            soname: 9
-          - version: "20221201"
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -7,19 +7,68 @@ on:
...
+          path: gems
+
   compile-and-test-system-dependencies:
+    name: Ruby ${{ matrix.ruby }} - libre2.${{ matrix.libre2.soname }}
     needs: ["build-cruby-gem"]
-    name: System Dependencies - ${{ matrix.sys }} vendored libs - Ruby ${{ matrix.ruby }} - libre2 ABI version ${{ matrix.libre2.soname }}
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        sys: ["enable", "disable"]
```
mudge, line 22: @stanhu do you have any thoughts about pulling this seeing as we have `compile-and-test-vendored-dependencies` which _should_ cover the same cases (compiling the gem and its vendored dependencies on all supported platforms)?

stanhu: As I recall, I added this in https://github.com/mudge/re2/pull/93 which covers the case when system libraries are installed, but the gem wants to use the vendored libraries and compile from source.

Does one of the checks in https://github.com/mudge/re2/actions/runs/6253089644/job/16977764503 cover this case?

mudge: Ah, let me add a specific test case for that: using the vendored libraries but with libre2 installed in Ruby's `exec_prefix` (I guess we don't need to exhaustively test that against every platform and Ruby version?)

mudge: I've added a dedicated job for testing the installation of the gem (with the default `---disable-system-libraries`) but where there is already a libre2 in the Ruby `exec_prefix` (and a system version of Abseil installed).

mudge: I've updated this to use whatever `libre2-dev` is available on `ubuntu-latest` but still symlinking it into Ruby's `lib` directory.

mudge:
@stanhu this shaves a significant amount of time from a build down (down from over 30 minutes in some cases to ~15 minutes): are you happy we're not losing any important coverage by testing installing the gem (with and without `--enable-system-libraries`) rather than testing `rake compile spec` directly?

.github/workflows/tests.yml
stanhu: Sounds good.

stanhu:
>  are you happy we're not losing any important coverage by testing installing the gem (with and without --enable-system-libraries) rather than testing rake compile spec directly?

Yes, that sounds good.

PR #106, opened 2023-09-23
Prevent overflow if RE2 error strings are too long

mudge:
As strncpy will truncate any error messages over 100 characters long without null termination, switch to using the safer strlcpy which guarantees null termination.


mudge:
We can probably add a test for this as RE2 will include the offending portion of an invalid pattern in its error message so writing a test with a 100+ character invalid pattern should br safely truncated.

PR #116, opened 2023-11-10
Modernize usage of Ruby's C extension API

casperisfine:
Use the `TypedData` API introduced in Ruby 1.9.2, instead of the old `Data` API that was deprecated in Ruby 2.3.

This allows to implement Write Barriers.

Write barrier protected objects are allowed to be promoted to the old generation, which means they only get marked on major GC, making minor GC faster.

The downside is that the `RB_BJ_WRITE` macro MUST be used to set references, otherwise the referenced object may be garbaged collected.

This also allow to implement GC compaction.

### DIFF
--- a/ext/re2/re2.cc
+++ b/ext/re2/re2.cc
@@ -183,12 +183,33 @@ static const rb_data_type_t re2_scanner_data_type = {
   .flags = RUBY_TYPED_FREE_IMMEDIATELY
 };
 
-static void re2_regexp_free(re2_pattern* self) {
+static void re2_regexp_free(void *data) {
+  re2_pattern* self = (re2_pattern*)data;
   if (self->pattern) {
     delete self->pattern;
   }
   free(self);
 }
+
+static size_t re2_regexp_memsize(const void *data) {
+  const re2_pattern* self = (const re2_pattern*)data;
+
+  size_t size = sizeof(re2_pattern);
+  if (self->pattern) {
+    size += sizeof(self->pattern);
+  }
+  return size;
+}
+
+static const rb_data_type_t re2_regexp_data_type = {
+  .wrap_struct_name = "RE2::Regexp",
+  .function = {
+    .dmark = NULL,
+    .dfree = re2_regexp_free,
+    .dsize = re2_regexp_memsize,
+  },
+  .flags = RUBY_TYPED_FREE_IMMEDIATELY
+};
 
 static VALUE re2_matchdata_allocate(VALUE klass) {
   re2_matchdata *m;
@@ -289,7 +310,7 @@ static VALUE re2_scanner_scan(VALUE self) {
   re2_scanner *c;
 
   TypedData_Get_Struct(self, re2_scanner, &re2_scanner_data_type, c);
-  Data_Get_Struct(c->regexp, re2_pattern, p);
+  TypedData_Get_Struct(c->regexp, re2_pattern, &re2_regexp_data_type, p);
 
   std::vector<RE2::Arg> argv(c->number_of_capturing_groups);
   std::vector<RE2::Arg*> args(c->number_of_capturing_groups);
@@ -345,7 +366,7 @@ static re2::StringPiece *re2_matchdata_find_match(VALUE idx, const VALUE self) {
   re2_pattern *p;
 
   TypedData_Get_Struct(self, re2_matchdata, &re2_matchdata_data_type, m);
-  Data_Get_Struct(m->regexp, re2_pattern, p);
+  TypedData_Get_Struct(m->regexp, re2_pattern, &re2_regexp_data_type, p);
 
   int id;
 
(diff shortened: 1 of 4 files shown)


ext/re2/re2.cc
```diff
@@ -122,18 +122,34 @@ static void parse_re2_options(RE2::Options* re2_options, const VALUE options) {
...
   rb_gc_mark(self->text);
 }
 
-static void re2_matchdata_free(re2_matchdata* self) {
+static void re2_matchdata_free(void * data) {
+  re2_matchdata* self = (re2_matchdata*)data;
   if (self->matches) {
     delete[] self->matches;
   }
   free(self);
```
mudge, line 151: Should we be using `xfree` here (and all other dfree functions) as per @peterzhu2118’s https://blog.peterzhu.ca/ruby-c-ext-part-7/#dfree?

ext/re2/re2.cc
```diff
@@ -122,18 +122,34 @@ static void parse_re2_options(RE2::Options* re2_options, const VALUE options) {
...
 }
 
+static size_t re2_matchdata_memsize(const void * data) {
+  return sizeof(re2_matchdata);
```
mudge, line 155: Should we be adding the size of the values in `data->matches` here too?

mudge (commented):
Thanks so much for contributing this, I was unaware of the newer `TypedData` API.

ext/re2/re2.cc
```diff
@@ -1358,9 +1364,11 @@ static VALUE re2_regexp_match(int argc, VALUE *argv, const VALUE self) {
     VALUE matchdata = rb_class_new_instance(0, 0, re2_cMatchData);
     TypedData_Get_Struct(matchdata, re2_matchdata, &re2_matchdata_data_type, m);
     m->matches = new(std::nothrow) re2::StringPiece[n];
-    m->regexp = self;
-    m->text = rb_str_dup(text);
-    rb_str_freeze(m->text);
+    RB_OBJ_WRITE(matchdata, &m->regexp, self);
+    if (!RTEST(rb_obj_frozen_p(text))) {
```
mudge, line 1391: If we don’t have a test for this already, we should add coverage for passing both a frozen and unfrozen string to `match`.

mudge: Added in 6cc08657ebdc4b03e5ba03590dc34ab6d5f6579d.

ext/re2/re2.cc
```diff
@@ -122,49 +122,135 @@ static void parse_re2_options(RE2::Options* re2_options, const VALUE options) {
...
+};
+
+static void re2_scanner_mark(void *data) {
+  re2_scanner* self = (re2_scanner *)data;
+  rb_gc_mark_movable(self->regexp);
+  rb_gc_mark_movable(self->text);
+}
+
+static void re2_scanner_update_references(void * data) {
+  re2_matchdata* self = (re2_scanner*)data;
```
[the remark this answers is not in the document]
mudge: Fixed in 772759d8fa8dec4c48f129e84ab04ff2d38064ca.

ext/re2/re2.cc
mudge: Fixed in 0bb9b0de7c9def5b7b439f0981ba6dbeb199245f.

ext/re2/re2.cc
mudge: Fixed in 8ded5101f913b5b29428916f7c57b6e3a6fa68b6.

mudge:
I also ran this through [ruby_memcheck](https://github.com/Shopify/ruby_memcheck) and no leaks were reported.

ext/re2/re2.cc
byroot: Actually, I think it should be `sizeof(re2::StringPiece) * self->number_of_matches`.

Not a huge deal though, the memsize is only for `ObjectSpace.memsize_of` and `ObjectSpace.dump`.

ext/re2/re2.cc
byroot: Yeah sorry, shouldn't have been in the PR at all, sometimes I just can't help myself to add code to avoid copies :)

ext/re2/re2.cc
mudge: I amended this in https://github.com/mudge/re2/commit/677352de905bd60f5616b77ffeeee8c4098eefa5 as I noticed the reported `memsize` from `ObjectSpace.dump` didn't change when there were more matches but that uses `sizeof(self->matches)` rather than `sizeof(re2::StringPiece)` as the basis of the calculation. I'll update it to use your suggestion as that seems slightly more accurate (though, as you say, this isn't a show-stopper).

mudge: Updated (once again) in https://github.com/mudge/re2/commit/9cdb8a900d2fabd6d071055e3f1ecbd99698d114.

mudge: @byroot does this mean we should also use `sizeof(*s->input)` rather than `sizeof(s->input)` to use the size of the `StringPiece` rather than the size of the _pointer_ to a `StringPiece` (same for `sizeof(p->pattern)` and `sizeof(s->set)` where they are both pointers)?

casperisfine: Yes, indeed. (Sorry, I'm a terrible C developer, and even worse C++ dev).

casperisfine: Basically, you should to account for all the structs that were `malloc` or otherwise allocated by that Object.

PR #117, opened 2023-11-22
Fix "non-trivial designated initializers not supported" errors

stanhu:
With g++ v7.3.1 (used with Amazon Linux 2), compiling the extension would fail with these errors:

```
compiling re2.cc
re2.cc:176:1: sorry, unimplemented: non-trivial designated initializers not
supported
 };
 ^
re2.cc:221:1: sorry, unimplemented: non-trivial designated initializers not
supported
 };
 ^
re2.cc:251:1: sorry, unimplemented: non-trivial designated initializers not
supported
 };
 ^
re2.cc:1610:1: sorry, unimplemented: non-trivial designated initializers not
supported
 };
 ^
```

This appears to be happening because the `parent` and `data` fields in `rb_data_type_struct` were omitted.

According to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55606, this issue was fixed in GCC 8.

### DIFF
--- a/ext/re2/re2.cc
+++ b/ext/re2/re2.cc
@@ -124,7 +124,7 @@ static void parse_re2_options(RE2::Options* re2_options, const VALUE options) {
 
 /* For compatibility with ruby < 2.7 */
 #ifdef HAVE_RB_GC_MARK_MOVABLE
-#define re2_compact_callback(x) .dcompact = (x),
+#define re2_compact_callback(x) (x),
 #else
 #define rb_gc_mark_movable(x) rb_gc_mark(x)
 #define re2_compact_callback(x)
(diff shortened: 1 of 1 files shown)


mudge:
Thanks for raising this, @stanhu. Am I right in thinking support for designated initialisers in C++ is a bit uncertain? Would we instead be better using the more "boring" form with only ordered expressions ala https://github.com/ruby/ruby/blob/60e19a0b5fc9c067ee88751192dc56da618f5060/weakmap.c#L151-L160?

For reference, here are the two versions of `rb_data_type_t` we need to support:

* [Ruby 2.6](https://github.com/ruby/ruby/blob/69f9992ed41920389d4185141a14f02f89a4d306/include/ruby/ruby.h#L1136-L1149)
* [Ruby >2.7](https://github.com/ruby/ruby/blob/1f4d4558484b370999954f3ede7e3aa3a3a01ef3/include/ruby/ruby.h#L1148-L1162)

mudge:
@stanhu I've pushed a commit to switch away from designated initializers: can you please test this and let me know if it works with g++ 7.3.1?

stanhu:
Sounds good, thanks!

mudge:
I've now published [2.4.3](https://rubygems.org/gems/re2/versions/2.4.3) to fix this. Please let me know if there are any other issues!

mudge:
@byroot this issue and the two potential fixes (see the two commits) may be of interest to you if you’re trying to convert other C++ extensions from `Data` to `TypedData`.

byroot:
Thanks for the ping, this is good to know. But hopefully there isn't too many of those around 😄

PR #118, opened 2023-11-24

stanhu:
The linker order problem in  abseil/abseil-cpp#1497 was shipped with abseil-cpp 20230802.0.

pkgconf v2.1.0 fixed pkgconf/pkgconf#268 and pkgconf/pkgconf#322, so we should no longer need to add the abseil linker flags manually.

Update CI to use pkgconf v2.1.0 for Windows machines and a note in README.me about using the right version.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -144,6 +144,8 @@ jobs:
         with:
           name: cruby-gem
           path: gems
+      - name: Show pkg-config version
+        run: pkg-config --version
       - run: ./scripts/test-gem-install gems
         shell: bash
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ If you're using Bundler, you can use the
 [`force_ruby_platform`](https://bundler.io/v2.3/man/gemfile.5.html#FORCE_RUBY_PLATFORM)
 option in your Gemfile.
 
+Windows users attempting to compile [abseil] must use pkgconf 2.1.0 or
+later, or builds will fail with [`undefined reference` errors](https://github.com/pkgconf/pkgconf/issues/322):
+
+    pacman -Sy mingw64/mingw-w64-x86_64-pkgconf
+
+This is not needed when using the precompiled gem or building against a system RE2 library.
+
 Documentation
 -------------
 
--- a/ext/re2/extconf.rb
+++ b/ext/re2/extconf.rb
@@ -271,65 +271,6 @@ def build_with_system_libraries
   build_extension
 end
 
-# pkgconf v1.9.3 on Windows incorrectly sorts the output of `pkg-config
-# --libs --static`, resulting in build failures: https://github.com/pkgconf/pkgconf/issues/268.
-# To work around the issue, store the correct order of abseil flags here and add them manually
-# for Windows.
-#
-# Note that `-ldbghelp` is incorrectly added before `-labsl_symbolize` in abseil:
-# https://github.com/abseil/abseil-cpp/issues/1497
-ABSL_LDFLAGS = %w[
-  -labsl_flags
-  -labsl_flags_internal
-  -labsl_flags_marshalling
-  -labsl_flags_reflection
-  -labsl_flags_private_handle_accessor
-  -labsl_flags_commandlineflag
-  -labsl_flags_commandlineflag_internal
-  -labsl_flags_config
-  -labsl_flags_program_name
-  -labsl_cord
-  -labsl_cordz_info
-  -labsl_cord_internal
-  -labsl_cordz_functions
-  -labsl_cordz_handle
-  -labsl_crc_cord_state
-  -labsl_crc32c
-  -labsl_crc_internal
-  -labsl_crc_cpu_detect
-  -labsl_raw_hash_set
-  -labsl_hash
-  -labsl_city
-  -labsl_bad_variant_access
-  -labsl_low_level_hash
-  -labsl_hashtablez_sampler
-  -labsl_exponential_biased
-  -labsl_bad_optional_access
-  -labsl_str_format_internal
-  -labsl_synchronization
-  -labsl_graphcycles_internal
-  -labsl_kernel_timeout_internal
-  -labsl_stacktrace
-  -labsl_symbolize
-  -ldbghelp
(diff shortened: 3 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -144,6 +146,9 @@ jobs:
         with:
           name: cruby-gem
           path: gems
+      - name: Update pkgconf
+        if: env.MATRIX_RUNS_ON == 'windows-2019' || env.MATRIX_RUNS_ON == 'windows-2022'
+        run: pacman -Sy --noconfirm pkgconf
```
stanhu, line 151: https://github.com/mudge/re2/actions/runs/6980696994/job/18996462142 shows that v2.0.2 is being used. Watching https://packages.msys2.org/base/pkgconf for the update.

.github/workflows/tests.yml
```diff
@@ -144,6 +146,9 @@ jobs:
         with:
           name: cruby-gem
           path: gems
+      - name: Update pkgconf
+        if: env.MATRIX_RUNS_ON == 'windows-2019' || env.MATRIX_RUNS_ON == 'windows-2022'
+        run: pacman -Sy --noconfirm pkgconf mingw64/mingw-w64-x86_64-pkgconf
```
stanhu, line 151: I was puzzled at first why https://github.com/mudge/re2/actions/runs/7039787062/job/19159476634#step:6:190 was failing. Adding debugging showed the wrong version of `pkgconf` was installed:

```
=== pkg-config version: 2.0.3
```

Then I noticed https://github.com/mudge/re2/actions/runs/7039660736/job/19159114189 passed successfully since 2.1.0 was installed, and then realized `mingw64/mingw-w64-x86_64-pkgconf` was the actual binary needed by Windows here.

README.md
```diff
@@ -62,6 +62,9 @@ If you're using Bundler, you can use the
 [`force_ruby_platform`](https://bundler.io/v2.3/man/gemfile.5.html#FORCE_RUBY_PLATFORM)
 option in your Gemfile.
 
+Note that Windows builds may fail with [`undefined reference` errors](https://github.com/pkgconf/pkgconf/issues/322)
```
mudge, line 65: Does this only affect people who are trying to compile Abseil on Windows (so users of the precompiled gem or anyone who is compiling against a system RE2 are unaffected)?

stanhu: Yes, I added a clarifying statement.

.github/workflows/tests.yml
mudge: I’m unfamiliar with `pacman` but are you installing two packages here? One called [`pkgconf`](https://packages.msys2.org/package/pkgconf) and one called [`mingw64/mingw-w64-x86_64-pkgconf`](https://packages.msys2.org/package/mingw-w64-x86_64-pkgconf?repo=mingw64)? What’s the difference? Is one a binary vs one that compiles from source?

.github/workflows/tests.yml
```diff
@@ -134,6 +134,8 @@ jobs:
             runs-on: "windows-2022"
           - ruby: "3.2"
             runs-on: "windows-2022"
+    env:
```
mudge, line 137: Minor thing but could this also be expressed in the matrix above as a new variable, e.g. `requires-pkgcon-upgrade: true` on the various `windows` entries and then you can simplify the `if` in the job?

.github/workflows/tests.yml
stanhu: Yes, there are two packages. Likely we only need the `mingw64` version, but I included the other one just in case. https://www.msys2.org/wiki/MSYS2-introduction/ has an explanation:

> MSYS2 consists of three subsystems and their corresponding package repositories, msys2, mingw32, and mingw64.
>
> The mingw subsystems provide native Windows programs and are the main focus of the project. These programs are built to co-operate well with other Windows programs, independently of the other subsystems. This part builds on the MinGW-w64 project.
>
> The msys2 subsystem provides an emulated mostly-POSIX-compliant environment for building software, package management, and shell scripting. These programs live in a virtual single-root filesystem (the root is the MSYS2 installation directory). Some effort is made to have the programs work well with native Windows programs, but it's not seamless. This part builds on the Cygwin project.

.github/workflows/tests.yml
stanhu: Let's try that.

.github/workflows/tests.yml
stanhu: I just dropped the `pkgconf` entry since abseil doesn't build with Cygwin.

stanhu: For some reason, Windows 2022 builds don't appear to have `pkgconf` installed (https://github.com/mudge/re2/actions/runs/7052548164/job/19197882051?pr=118), unlike Windows 2019 (https://github.com/mudge/re2/actions/runs/7052548164/job/19197881418?pr=118). 

My only guess that Windows 2022 ships with the standard `pkg-config`, which doesn't have the bugs in `pkgconf`. https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md doesn't really show much difference from https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md, though.

mudge: Does this mean we don’t need it for windows-2022 builds or do we depend on pkgconf (rather than pkg-config) for Windows?

stanhu: I'm not sure. I wish there were a way to inspect the GitHub VMs because it's really opaque to me what's actually installed. https://github.com/msys2/MSYS2-packages/issues/2872 claims that you can't install `pkgconf` if `pkg-config` is already installed. I'll try to drop it from Windows 2022 and see what happens.

mudge: It looks like there was an update to the Windows 2019 image in the past 24 hours: https://github.com/actions/runner-images/commit/45a5b970b809e08869dcf5f11456001d86952d3f

stanhu: Yeah, the latest builds seem to work fine without the `pkgconf` upgrade now. I added a `pkg-config --version` check to make it easier to triage if the problem happens again.

mudge (approved):

Issue #158, opened 2024-07-14
mini_portile2 integration tests are failing


flavorjones:
Heyo @mudge, I need to ask for your help to diagnose a thing.

I've been running downstream integration tests in mini_portile2's Github Actions suite for a while, and this week the re2 tests started failing and I don't really understand the failure mode, which appears to be in building abseil.

Example: https://github.com/flavorjones/mini_portile/actions/runs/9903843042/job/27429080214#step:7:60

```
---------- IMPORTANT NOTICE ----------
Building re2 with a packaged version of abseil-20240116.2.
Configuration options: -DCMAKE_SYSTEM_NAME\=Linux -DCMAKE_SYSTEM_PROCESSOR\=x86_64 -DCMAKE_C_COMPILER\=gcc -DCMAKE_CXX_COMPILER\=g++ -DCMAKE_BUILD_TYPE\=Release -DCMAKE_CXX_STANDARD\=14 -DCMAKE_POSITION_INDEPENDENT_CODE\=ON -DCMAKE_INSTALL_LIBDIR\=lib -DCMAKE_CXX_VISIBILITY_PRESET\=hidden -DABSL_PROPAGATE_CXX_STD\=ON
Extracting 20240116.2.tar.gz into tmp/x86_64-pc-linux/ports/abseil/20240116.2... OK
Running 'configure' for abseil 20240116.2... ERROR. Please review logs to see what happened:
----- contents of '/tmp/d20240714-1847-650jmm/tmp/x86_64-pc-linux/ports/abseil/20240116.2/configure.log' -----
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test ABSL_INTERNAL_AT_LEAST_CXX17
-- Performing Test ABSL_INTERNAL_AT_LEAST_CXX17 - Failed
-- Performing Test ABSL_INTERNAL_AT_LEAST_CXX20
-- Performing Test ABSL_INTERNAL_AT_LEAST_CXX20 - Failed
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Configuring done (1.9s)
[...]

PR #159, opened 2024-07-16
Add patch to fix building Abseil with CMake 3.30.0

mudge:
Fixes https://github.com/mudge/re2/issues/158

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -16,6 +16,9 @@ re2_archive = File.join("ports/archives", File.basename(re2_recipe.files[0][:url
 
 re2_gemspec.files << abseil_archive
 re2_gemspec.files << re2_archive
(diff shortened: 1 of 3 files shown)


stanhu:
Looks good to me, but Ruby 2.6 is failing for some reason: https://github.com/mudge/re2/actions/runs/9952702253/job/27494586233?pr=159:

mudge:
I suspect the problem is that the issue here is the custom [libre2-dev](https://github.com/mudge/re2-ci/releases/tag/v2) deb, let me try building a new one against the latest Abseil.

mudge:
Fixed by rebuilding `libre2-dev_20230701_amd64.deb` against the latest version of Abseil.

PR #167, opened 2025-01-20
Set RUBY_CC_VERSION using RakeCompilerDock.set_ruby_cc_version

stanhu:
This avoids the need to hard-code specific patch-level versions.

See https://github.com/rake-compiler/rake-compiler-dock/releases/tag/v1.9.1.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -335,7 +335,7 @@ jobs:
           docker run --rm -v "$(pwd):/re2" -w /re2 \
             --platform=linux/arm64 \
             ruby:${{ matrix.ruby }}-alpine \
-            ./scripts/test-gem-install ${{ matrix.rubygems }}
+            /bin/sh -c "apk update && apk add libstdc++ && ./scripts/test-gem-install ${{ matrix.rubygems }}"
 
   test-precompiled-arm-linux-gnu:
     needs: "precompile-arm-linux-gnu"
@@ -393,7 +393,7 @@ jobs:
           docker run --rm -v "$(pwd):/re2" -w /re2 \
             --platform=linux/arm/v7 \
             ruby:${{ matrix.ruby }}-alpine \
-            ./scripts/test-gem-install ${{ matrix.rubygems }}
+            /bin/sh -c "apk update && apk add libstdc++ && ./scripts/test-gem-install ${{ matrix.rubygems }}"
 
   test-precompiled-x86-linux-gnu:
     needs: "precompile-x86-linux-gnu"
@@ -451,7 +451,7 @@ jobs:
           docker run --rm -v "$(pwd):/re2" -w /re2 \
             --platform=linux/386 \
             ruby:${{ matrix.ruby }}-alpine \
-            ./scripts/test-gem-install ${{ matrix.rubygems }}
+            /bin/sh -c "apk update && apk add libstdc++ && ./scripts/test-gem-install ${{ matrix.rubygems }}"
 
   test-precompiled-x86_64-linux-gnu:
     needs: "precompile-x86_64-linux-gnu"
@@ -506,6 +506,7 @@ jobs:
         with:
           name: cruby-x86_64-linux-musl-gem
           path: pkg
+      - run: apk update && apk add libstdc++
       - run: ./scripts/test-gem-install ${{ matrix.rubygems }}
 
   test-precompiled-arm64-darwin:
--- a/Rakefile
+++ b/Rakefile
@@ -33,7 +33,7 @@ cross_platforms = %w[
   x86_64-linux-musl
 ].freeze
 
-ENV['RUBY_CC_VERSION'] = '3.4.1:3.3.5:3.2.6:3.1.6:3.0.7:2.7.8:2.6.10'
+ENV['RUBY_CC_VERSION'] = RakeCompilerDock.ruby_cc_version("3.4", "3.3", "3.2", "3.1", "3.0", "2.7", "2.6")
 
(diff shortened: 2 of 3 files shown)


mudge:
Precompilation seems to be failing in the vast majority of cases with the following error:

```
/usr/local/rbenv/versions/3.4.1/lib/ruby/3.4.0/psych/visitors/to_ruby.rb:388:in 'Psych::Visitors::ToRuby#deduplicate': undefined method 'untaint' for an instance of String (NoMethodError)
```

Have you seen this error elsewhere (given the only change here is the rake-compiler-dock version)?

cc @flavorjones

flavorjones:
> Have you seen this error elsewhere

No! I've upgraded nokogiri, sqlite3-ruby, and ruby-c-extensions-explained with no problem. And the rake-compiler-dock suite (which has its own test gem) is green.

> Somehow RUBY_VERSION is < 2.7, but the stacktrace shows this is Ruby 3.4.1

I bet this is coming from the reset in [rake-compiler's "fake.rb" rbconfig](https://github.com/rake-compiler/rake-compiler/blob/150c8e359748a4e7582daea6ecbe7ecb063d3086/lib/rake/extensiontask.rb#L571). Without spending more time I can't be sure what combination of things is causing this issue, especially since I haven't changed how we build the 2.7.8 cross ruby.

Maybe it's because the default Ruby in the container is now 3.4.1? Maybe you can try to run `rbenv shell 3.1` in your `RakeCompilerDock.sh` script to determine if that works around it.

flavorjones:
@mudge Also, and I know you don't want to hear it, but this is the kind of problem that tends to come up when building against older versions of Ruby. AKA the "too much effort" problem.[^1]

Don't get me wrong, ideally this _should_ work, and if I had an infinite amount of time I would absolutely continue to make old rubies work perfectly even as we make changes to accommodate new rubies. But I don't have that kind of time, and in my opinion it's not worth anyone's effort to precompile EOL rubies in native gems. See https://github.com/rake-compiler/rake-compiler-dock/issues/77

[^1]: https://github.com/rake-compiler/rake-compiler-dock/issues/134#issuecomment-2571645343

mudge:
Using `rbenv shell 3.1.6` seems to fix the issue.

stanhu:
>  But I don't have that kind of time, and in my opinion it's not worth anyone's effort to precompile EOL rubies in native gems. 

I agree with this. It also makes the precompiled gems a lot bigger.

flavorjones:
@mudge 

> now the musl builds are broken with a missing libstdc++

I don't think this is related to the rake-compiler-dock changes. The most recent ruby:alpine images have had a few of the `-dev` packages removed which may have caused this library file to also be removed.

stanhu:
There appears to be some inconsistency:

#### `ruby:3.4-alpine`

```
$ docker run -it ruby:3.4-alpine sh
/ # ls -al /usr/lib/libstd*
lrwxrwxrwx    1 root     root            19 Jan 15 17:51 /usr/lib/libstdc++.so.6 -> libstdc++.so.6.0.33
-rwxr-xr-x    1 root     root       2771336 Oct 17 20:43 /usr/lib/libstdc++.so.6.0.33
/ # apk info --who-owns /usr/lib/libstdc++.so.6
/usr/lib/libstdc++.so.6 is owned by libstdc++-14.2.0-r4
/ # apk list libstdc++
WARNING: opening from cache https://dl-cdn.alpinelinux.org/alpine/v3.21/main: No such file or directory
WARNING: opening from cache https://dl-cdn.alpinelinux.org/alpine/v3.21/community: No such file or directory
libstdc++-14.2.0-r4 x86_64 {gcc} (GPL-2.0-or-later AND LGPL-2.1-or-later) [installed]
```

#### `ruby:3.3-alpine`

```
$ docker run -it ruby:3.3-alpine sh
/ # ls -al /usr/lib/libstd*
ls: /usr/lib/libstd*: No such file or directory
/ # apk list libstdc++
WARNING: opening from cache https://dl-cdn.alpinelinux.org/alpine/v3.21/main: No such file or directory
WARNING: opening from cache https://dl-cdn.alpinelinux.org/alpine/v3.21/community: No such file or directory
```

It looks like running `apk update && apt add libstdc++` solves the problem.

mudge (approved):

flavorjones:
I added a small feature to make it easier to choose the ruby being used, and would appreciate any thoughts on it: https://github.com/rake-compiler/rake-compiler-dock/pull/151

stanhu:
> Maybe it's because the default Ruby in the container is now 3.4.1? Maybe you can try to run rbenv shell 3.1 in your RakeCompilerDock.sh script to determine if that works around it.

@flavorjones That looks like it worked here. But it seems that fake `RUBY_VERSION` could cause all sorts of weird bugs in the future, even for supported Ruby versions. Right now I mostly see `RUBY_VERSION < 2.7` in the Ruby code base, but if someone introduces a Ruby version-specific feature, then we might hit a similar issue again due to the mismatch.

I don't fully understand https://github.com/tjschuck/rake-compiler-dev-box/issues/5, but I wonder if this workaround is still needed. I should hope most gem authors don't ever need to know about https://github.com/rake-compiler/rake-compiler-dock/pull/151. 😄

flavorjones:
@stanhu To be clear, the comment referencing the box project is referring to the 'resolv' require, not the fake_rb method. I believe that fake_rb is fundamental to how rake-compiler works, but I also don't fully understand rake-compiler.

Maybe this is an opportunity to open an issue on rake-compiler to talk about this use case and the errors we saw in the build?

PR #169, opened 2025-02-03
Keep rake_compiler_dock as a development dependency

stanhu:
rake-compiler-dock is only needed to build native gems. For Ruby platform installs, we should not need this.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'rake/extensiontask'
-require 'rake_compiler_dock'
 require 'rspec/core/rake_task'
 
 require_relative 'ext/re2/recipes'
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -63,6 +61,9 @@ end
 
 namespace :gem do
   cross_platforms.each do |platform|
+    require 'rake_compiler_dock'
```
mudge, line 64: Don't we require rake-compiler-dock at the top of this file as well?

Is it fair to say the entire Rakefile is a development-only concern? It's included in the gemspec (presumably because we need it for CI) but is it actually loaded by an end-user of the gem?

stanhu: > Is it fair to say the entire Rakefile is a development-only concern? It's included in the gemspec (presumably because we need it for CI) but is it actually loaded by an end-user of the gem?

Yeah, I think that's true. For some reason I thought the `Rakefile` might be needed when installing `--platform ruby`.

PR #173, opened 2025-06-28
Upgrade to Abseil 20250512.1

mudge:
- **Upgrade Abseil to 20250512.1**
- **Check for new versions of Abseil again**


mudge:
This requires compiling the macOS gems for macOS 10.14.

mudge:
Let's leave this until GitHub Actions removes the macos-13 runners as a good sign that support for macOS High Sierra is no longer tenable.

